PR #13, opened 2018-02-28
chore: added .env file to the .gitignore

MatousJobanek:
added .env file to the .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ oauth.token
 
 # Generated files
 cluster/generated/
+
+# autoenv config file
+.env


bartoszmajsak:
I don't think we should add it here, as it's strictly user preference / tool of choice. For the same reason, I don't have `.autoenv.zsh` in `.gitignore`. What I do have instead is global `.gitignore` which works for all my projects.

`git config --global core.excludesfile ~/.gitignore`

If we agree that's the better approach I can add it to the documentation.

MatousJobanek:
Agree, I use the global .gitignore file as well, but I completely forgot to use it now... 
Thx for pointing this out

Issue #3, opened 2018-02-26
Development documentation


bartoszmajsak:
Provide documentation describing development process

- [x] build
- [x] local setup
- [x] deployment
- [x] testing real GH hooks

PR #14, opened 2018-02-28
fix(#3): improves documentation

bartoszmajsak:
- better explanation about GitHub webhooks
- plugin configuration section


bartoszmajsak:
Thanks, haven't noticed that. We need to make a plugin generating per-pr doc preview (based on travis scripts we have).

PR #15, opened 2018-03-01
chore: moved the common plugin logic from mains to separated function

MatousJobanek:
There was a lot of duplications in the main functions of the plugins - I moved it to a separated function in a separated file.
If I used a wrong abstraction, approach, file/function names or the place for the file, then just tell me. My main problem is that I'm trying to write Java in Go - I need to learn the Go-way properly.

### DIFF
--- a/plugin/plugin_init.go
+++ b/plugin/plugin_init.go
@@ -28,12 +28,12 @@ var (
 	webhookSecretFile = flag.String("hmac-secret-file", "/etc/webhook/hmac", "Path to the file containing the GitHub HMAC secret.")
 )
 
+type EventHandlerCreator func(client *github.Client) server.GitHubEventHandler
+type ServerCreator func(hmacSecret []byte, testKeeperEvents server.GitHubEventHandler) *server.Server
+
 // InitPlugin instantiates logger, loads the secrets from the flags, sets context to background and starts server with
 // the attached event handler.
-func InitPlugin(
-	log *logrus.Entry,
-	createHandler func(Client *github.Client) server.GitHubEventHandler,
-	createServer func(HmacSecret []byte, testKeeperEvents server.GitHubEventHandler) *server.Server,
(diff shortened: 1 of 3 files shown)


plugin/plugin_init.go
```diff
@@ -0,0 +1,75 @@
...
+	githubTokenFile   = flag.String("github-token-file", "/etc/github/oauth", "Path to the file containing the GitHub OAuth secret.")
+	webhookSecretFile = flag.String("hmac-secret-file", "/etc/webhook/hmac", "Path to the file containing the GitHub HMAC secret.")
+)
+
+// InitPlugin instantiates logger, loads the secrets from the flags, sets context to background and starts server with
+// the attached event handler.
+func InitPlugin(
+	log *logrus.Entry,
+	createHandler func(Client *github.Client) server.GitHubEventHandler,
+	createServer func(HmacSecret []byte, testKeeperEvents server.GitHubEventHandler) *server.Server,
```
bartoszmajsak: We can make these functions types like `externalplugins.ExternalPluginHelpProvider` is. WDYT?

plugin/pr-sanitizer/main.go
```diff
@@ -1,35 +1,18 @@
...
-
 	"github.com/sirupsen/logrus"
 	"github.com/google/go-github/github"
-	"k8s.io/test-infra/prow/pluginhelp/externalplugins"
-	"k8s.io/test-infra/prow/plugins"
 	"k8s.io/test-infra/prow/pluginhelp"
 
 	"github.com/arquillian/ike-prow-plugins/plugin/server"
-	. "github.com/arquillian/ike-prow-plugins/plugin/utils"
+	commonPlugin "github.com/arquillian/ike-prow-plugins/plugin"
```
bartoszmajsak: I would call it `pluginBootstrap` or something. `common` is a bit too ambiguous.

plugin/plugin_init.go
MatousJobanek: good point :1st_place_medal:

Issue #2, opened 2018-02-26
Implement tests for existing logic


bartoszmajsak:
- unit tests for event handling
- integration/sv tests for PR updates

bartoszmajsak:
Test-after, but well... better late then never ;)

PR #19, opened 2018-03-02
feat(#2): feature tests for test-keeper plugin

bartoszmajsak:
This PR brings `ginkgo` feature tests for status updates of Test Keeper plugin.


plugin/test-keeper/plugin/event_handler_test.go
```diff
@@ -0,0 +1,180 @@
...
+		})
+
+		It("should approve opened pull request when tests included", func() {
+			// given
+			gock.New("https://api.github.com").
+				Get("/repos/bartoszmajsak/wfswarm-booster-pipeline-test/pulls/2/files").
+				Reply(200).
+				Body(fromJson("test_fixtures/github_calls/prs/with_tests/changes.json"))
+
+			toHaveSuccessState := func(statusPayload map[string]interface{}) (bool) {
```
MatousJobanek: these functions `toHaveSuccessState` and `toHaveFailureState` can be as a global variables

plugin/test-keeper/plugin/event_handler_test.go
```diff
@@ -0,0 +1,180 @@
...
+			// given
+			gock.New("https://api.github.com").
+				Get("/repos/bartoszmajsak/wfswarm-booster-pipeline-test/pulls/1").
+				Reply(200).
+				Body(fromJson("test_fixtures/github_calls/prs/without_tests/pr_details.json"))
+
+			gock.New("https://api.github.com").
+				Get("/repos/bartoszmajsak/wfswarm-booster-pipeline-test/collaborators/bartoszmajsak-test/permission").
+				Reply(200).
+				Body(fromJson("test_fixtures/github_calls/collaborators_bartoszmajsak-test_permission.json"))
```
MatousJobanek: the files
`collaborators_bartoszmajsak-test_permission.json`
`collaborators_bartoszmajsak_permission.json`
have very similar names (which is very confusing) and it is not obvious what they represent. 
btw, firstly I read the first file as `collaborators_bartoszmajsak_test_permission.json` which confused me totally.
It would be good to rename the files (or the users) to have something more meaningful:
`bartoszmajsak-admin`
`bartoszmajsak-non-admin`
or maybe better:
`ike-admin`
`ike-non-admin` 
without the specific usernames.
The same applies to the rest of the files and calls. I know why you put it there - you just copy-pasted the responses, but it is quite hard to modify/reproduce it for the other users. Mainly because it is releated to your personal account, repository and PRs, that can by modified/removed by you anytime.

plugin/test-keeper/plugin/even_handler_suite_test.go
```diff
@@ -0,0 +1,15 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestTestKeeper(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Prow Event Handler Test Keeper Plugin Suite")
+	// TODO configure .gitignore or store in different folder
+	//junitReporter := reporters.NewJUnitReporter("junit.xml")
+	//RunSpecsWithDefaultAndCustomReporters(t, "Prow Event Handler Test Keeper Plugin Suite", []Reporter{junitReporter})
```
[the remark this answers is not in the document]
bartoszmajsak: Not really, I will make a new task for it. Originally I wanted to configure it as part of it, but this can wait.

plugin/test-keeper/plugin/event_handler.go
```diff
@@ -28,48 +32,48 @@ var (
...
+		)
+	}
 	switch eventType {
-	case "pull_request":
-		gh.log.Info("Pull request received")
+	case githubevents.PullRequest:
+		gh.Log.Info("Pull request received")
 		var event github.PullRequestEvent
 		if err := json.Unmarshal(payload, &event); err != nil {
+			gh.Log.Info("JSON: ", event)
```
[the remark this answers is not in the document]
bartoszmajsak: Fixed in e860a32. Lemme know if that's good enough.

plugin/test-keeper/plugin/event_handler_test.go
bartoszmajsak: Would it make the test more readable? You will have to jump between those. It's violating DRY, but keeps test's "story" in one place. I think about it as an assertion and like with those I tend not to extract it outside of the test.

bartoszmajsak: Are you fine with leaving it here or have strong preference on moving it to global variable?

MatousJobanek: I don't insist on it. Let's keep it there and if the file grows, then we can think about making it as a global variable to keep the file shorter.

plugin/test-keeper/plugin/even_handler_suite_test.go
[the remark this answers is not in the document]
bartoszmajsak: Removed in favor of #21

plugin/test-keeper/plugin/event_handler.go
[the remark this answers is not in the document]
MatousJobanek: Yeah, it's better. Just one thing - you could use `Infof` to get rid of the joining of the string using `+` char

plugin/test-keeper/plugin/event_handler_test.go
bartoszmajsak: As we discussed last week, long term goal would be to look at `go-vcr` or `hoverfly`, even though I like `gock` dsl for stubbing http calls. But even with that we will have to make recordings against real GH use case (users and repo).

I will rename the files to make it a bit less cryptic and let's think about overall improvement as we progress with the implementations.

plugin/test-keeper/plugin/event_handler.go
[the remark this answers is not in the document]
bartoszmajsak: I think `Errorf` would be even more appropriate. Fixed with a6ce97e. Thanks :)

plugin/test-keeper/plugin/event_handler_test.go
bartoszmajsak: agree

plugin/test-keeper/plugin/event_handler_test.go
bartoszmajsak: > The same applies to the rest of the files and calls. I know why you put it there - you just copy-pasted the responses, but it is quite hard to modify/reproduce it for the other users. Mainly because it is releated to your personal account, repository and PRs, that can by modified/removed by you anytime.

I'm open to better names, but keep in mind it's a stub at this point and it is unlikely to be changed any time.

I renamed permission `json`s in 9d80b46. I hope it's slightly better now.

bartoszmajsak:
Based on our discussion on Mattermost I added section about testing in the docs 077f85c.

bartoszmajsak:
Thanks @MatousJobanek.

PR #39, opened 2018-03-14
feat(#38): adds goimports to the toolchain

### LINKED ISSUE #38: Introduce autoformatting as build target
Add `goimports` tool which takes care of formatting and removing unused imports as a `Makefile` target.

bartoszmajsak:
Important to note: if imported packages are not named properly `goimports` will remove them. Thus I extended docs to highlight this problem.

> `goimports` does not correctly handle import of packages where package name differs from last part of the URL provided. For example `import "gopkg.in/h2non/gock.v1"` will be removed. To overcome this problem name the import `import gock "gopkg.in/h2non/gock.v1"`. 

Fixes #38

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,6 @@ clean: ## Removes binary, cache folder and docker images
 
 .PHONY: tools
 tools: ## Installs required go tools
-	@go get -u github.com/alecthomas/gometalinter
 	@go get -u github.com/alecthomas/gometalinter && gometalinter --install
 	@go get -u github.com/onsi/ginkgo/ginkgo
 	@go get -u github.com/onsi/gomega
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,10 @@ test:
 .PHONY: build
 build: compile test check
 
+.PHONY: imports ## Removes unneeded imports and formats source code
+imports:
+	@goimports -l -w pkg
+
 # Build configuration
 BUILD_TIME=$(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
 COMMIT=$(shell git rev-parse --short HEAD)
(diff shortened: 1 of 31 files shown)


Makefile
```diff
@@ -32,6 +32,14 @@ clean: ## Removes binary, cache folder and docker images
 	@rm -rf ${BINARY_DIR}
 	@rm -rf $(PLUGIN_DEPLOYMENTS_DIR)
 
+.PHONY: tools
+tools: ## Installs required go tools
+	@go get -u github.com/alecthomas/gometalinter
+	@go get -u github.com/alecthomas/gometalinter && gometalinter --install
```
MatousJobanek: why the getting of gometalinter `@go get -u github.com/alecthomas/gometalinter` is called twice?

Makefile
```diff
@@ -32,6 +32,14 @@ clean: ## Removes binary, cache folder and docker images
 	@rm -rf ${BINARY_DIR}
 	@rm -rf $(PLUGIN_DEPLOYMENTS_DIR)
 
+.PHONY: tools
+tools: ## Installs required go tools
```
MatousJobanek: is this target called by default or do I need to call it separately? I don't see the target as part of the build or any other target.
Same for the target "imports"

bartoszmajsak: There is no need to run `tools` once you get it set up (e.g. `setup.sh`). I would think about it as a separate/isolated step. In fact we could re-use it as well in `.travis.yml`.

The same for `imports` - we could make it part of the `build` target but this will result in adding more changes to what you have done on your own so that might be a problem (mainly b/c of the aforementioned `import` removal).

Makefile
bartoszmajsak: Copy-pastie :) Somehow missed that. Fix on the way

bartoszmajsak: Fixed with d7a909f.

Issue #37, opened 2018-03-13
Comment on the PR when no tests are included


bartoszmajsak:
Next to PR check we should add automated commented in case when no tests are found in the Pull Request. This comment should point the developer who opened the Pull Requests to resources on how the tests are written in a given project.

* Extend configuration with a link to a file (relative to root of the project/repository)
* Include is as a comment made by the bot when no tests are found (note: this can be plain text or GH flavored `.md`, but we shouldn't really validate that), but only if the configuration is provided
* The comment should not be duplicated when PR is updated

PR #40, opened 2018-03-14
feat(#37): comments on PR when no tests are included

MatousJobanek:
#### Comment service
Before it adds a comment it checks all present comments in the issue/pull-request. 
* If no comment with main plugins title "Ike Plugins" is found, then it adds a new comment with the main title, assignee mention, the test-keeper plugin title and the plugin commentMsg. 
* If a comment with the main title is found but missing the test-keeper plugin title, then it modifies the comment by adding the plugin title with the comment message. 
* If everything is present already, then it does nothing.

#### Comment message
Comment message is constructed based on the fact if a configuration is present. 
* If a config file is not present, then it shows default comment referencing to project documentation. 
* If config is present but missing custom comment message there then it shows default with reference to the config file. 
* If config file with custom comment message is present, then it shows the given custom message.

##### Screenshots:
Default with reference to the documetation:
image

Default with reference to config file:
image

Comment with custom message
image

Fixes: #37

### DIFF
--- a/docs/test-keeper.adoc
+++ b/docs/test-keeper.adoc
@@ -26,7 +26,7 @@ include::../pkg/plugin/test-keeper/plugin/test_fixtures/github_calls/prs/with_te
 <1> Defines set of regular expressions which will be used to match files changeset and determine if Pull Requests comes with any changed test files.
 <2> These regular expressions will be used against changeset in the Pull Requests to exclude files that don't have to be verified by any test. If only such files exists the check will be marked as "Success" as no tests are expected for such a PR.
 <3> Allows you to decide if you want to combine your patterns with predefined defaults (`true` by default).
-<4> Sets either relative path or absolute URL to a file that contains a message to be added to the Pull Request when no test is found (if nothing is set then a default message is used).
+<4> Sets either relative path or absolute URL to a file that contains a message to be added to the Pull Request when no test is found (if nothing is set then a link:https://github.com/arquillian/ike-prow-plugins/tree/master/pkg/plugin/test-keeper/plugin/comment_message.go[default message] is used).
 
 IMPORTANT: The configuration file is always loaded from the `HEAD` of the Pull Request.
 
--- a/docs/test-keeper.adoc
+++ b/docs/test-keeper.adoc
@@ -1,7 +1,7 @@
 == Test Keeper plugin
 
 This plugin can help you stick to the rule that every feature you ship comes with automated way of assuring it works - by using automated tests. 
-If it won't find any tests in the Pull Request, it will mark its check as **Failure**, or as **Success** otherwise (see the screenshots below):
+If it won't find any tests in the Pull Request, it adds a comment with a description and marks its check as **Failure**, or as **Success** otherwise (see the screenshots below):
 
 image::images/testkeeper-success.png[Success, title="Success status"]
 
@@ -26,6 +26,7 @@ include::../pkg/plugin/test-keeper/plugin/test_fixtures/github_calls/prs/with_te
 <1> Defines set of regular expressions which will be used to match files changeset and determine if Pull Requests comes with any changed test files.
 <2> These regular expressions will be used against changeset in the Pull Requests to exclude files that don't have to be verified by any test. If only such files exists the check will be marked as "Success" as no tests are expected for such a PR.
 <3> Allows you to decide if you want to combine your patterns with predefined defaults (`true` by default).
+<4> Sets either relative path or absolute URL to a file that contains a message to be added to the Pull Request when no test is found (if nothing is set then a default message is used).
 
 IMPORTANT: The configuration file is always loaded from the `HEAD` of the Pull Request.
 
--- a/docs/test-keeper.adoc
+++ b/docs/test-keeper.adoc
@@ -26,9 +26,9 @@ include::../pkg/plugin/test-keeper/plugin/test_fixtures/github_calls/prs/with_te
 <1> Defines set of regular expressions which will be used to match files changeset and determine if Pull Requests comes with any changed test files.
 <2> These regular expressions will be used against changeset in the Pull Requests to exclude files that don't have to be verified by any test. If only such files exists the check will be marked as "Success" as no tests are expected for such a PR.
 <3> Allows you to decide if you want to combine your patterns with predefined defaults (`true` by default).
-<4> Sets either relative path or absolute URL to a file that contains a message to be added to the Pull Request when no test is found (if nothing is set then a link:https://github.com/arquillian/ike-prow-plugins/tree/master/pkg/plugin/test-keeper/plugin/comment_message.go[default message] is used).
+<4> Sets either relative path or absolute URL to a file that contains a plugin hint to be added to the Pull Request when no test is found (if nothing is set then a link:https://github.com/arquillian/ike-prow-plugins/tree/master/pkg/plugin/test-keeper/plugin/comment_message.go[default message] is used).
 
-IMPORTANT: The configuration file is always loaded from the `HEAD` of the Pull Request.
+IMPORTANT: Both the configuration file and the file with plugin hint are always loaded from the `HEAD` of the Pull Request.
 
 ==== Default exclusions
 
--- a/pkg/github/raw_file_service.go
+++ b/pkg/github/raw_file_service.go
@@ -1,0 +1,24 @@
+package github
+
+import (
+	"fmt"
+
+	"github.com/arquillian/ike-prow-plugins/pkg/scm"
+	"github.com/arquillian/ike-prow-plugins/pkg/utils"
+)
+
+// RawFileService encapsulates retrieval of files in the given GitHub repository change
+type RawFileService struct {
+	Change scm.RepositoryChange
+}
+
+// GetRawFile retrieves raw file content on the given path from the related GitHub repository change
+func (s *RawFileService) GetRawFile(path string) ([]byte, bool, error) {
+	return utils.GetFileFromURL(s.GetRawFileURL(path))
+}
+
+// GetRawFileURL creates a url to the given path related to the GitHub repository change
+func (s *RawFileService) GetRawFileURL(path string) string {
+	return fmt.Sprintf("https://raw.githubusercontent.com/%s/%s/%s/%s",
+		s.Change.Owner, s.Change.RepoName, s.Change.Hash, path)
+}
--- a/pkg/github/raw_file_service.go
+++ b/pkg/github/raw_file_service.go
@@ -13,7 +13,7 @@ type RawFileService struct {
 }
 
 // GetRawFile retrieves raw file content on the given path from the related GitHub repository change
-func (s *RawFileService) GetRawFile(path string) ([]byte, bool, error) {
+func (s *RawFileService) GetRawFile(path string) ([]byte, error) {
 	return utils.GetFileFromURL(s.GetRawFileURL(path))
 }
 
--- a/pkg/plugin/comment_service.go
+++ b/pkg/plugin/comment_service.go
@@ -11,8 +11,11 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// IkePluginsTitle is a constant containing "Ike Plugins" title with markdown formatting
-const IkePluginsTitle = "### Ike Plugins"
+// PluginTitleTemplate is a constant template containing "Ike Plugins (name-of-plugin)" title with markdown formatting
+const (
+	PluginTitleTemplate     = "### Ike Plugins (%s)"
+	assigneeMentionTemplate = "@%s Please, pay attention to the following message:"
+)
 
 // CommentService is a struct managing plugin comments
 type CommentService struct {
@@ -49,10 +52,9 @@ func NewCommentService(client *github.Client, log *logrus.Entry, change scm.Repo
 	}
 }
 
-// PluginComment checks all present comments in the issue/pull-request. If no comment with IkePluginsTitle is found, then
-// it adds a new comment with IkePluginsTitle, assignee, the plugin tittle and the given commentMsg. If a comment with
-// IkePluginsTitle is found but missing the particular plugin title, then it modifies the comment by adding plugin title
-// with the given commentMsg. If everything is present already, then it does nothing.
+// PluginComment checks all present comments in the issue/pull-request. If no comment with PluginTitleTemplate
+// (with the related plugin) is found, then it adds a new comment with the plugin title, assignee mention
+// and the given commentMsg. If such a comment is present already, then it does nothing.
 func (s *CommentService) PluginComment(commentMsg string) error {
 
 	comments, _, err := s.client.Issues.
@@ -63,16 +65,13 @@ func (s *CommentService) PluginComment(commentMsg string) error {
 
 	for _, com := range comments {
 		content := *com.Body
-		if strings.HasPrefix(content, IkePluginsTitle) {
-			if strings.Contains(content, s.getPluginTitle()) {
-				return nil
-			}
-			return s.appendToComment(commentMsg, com)
+		if strings.HasPrefix(content, s.getPluginTitle()) {
+			return nil
 		}
 	}
 
 	comment := &github.IssueComment{
-		Body: s.append(s.createBeginning(), s.createPluginParagraph(commentMsg)),
+		Body: s.createPluginHint(commentMsg),
 	}
 
 	_, _, err = s.client.Issues.CreateComment(context.Background(), s.issue.Owner, s.issue.RepoName, s.issue.Number, comment)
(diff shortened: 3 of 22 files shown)


MatousJobanek:
I added a commit that changes the config parameters to `comment_message_file` and the custom messages from the given location - can be either absolute URL or relative path.
if the file on the given location is not found (or is any other problem with loading it), then the message contains a mention of it:
image
This is, however, something that I wasn't sure about - if I should only log it (send to the Sentry) or add it to the comment... WDYT?

docs/test-keeper.adoc
```diff
@@ -26,6 +26,7 @@ include::../pkg/plugin/test-keeper/plugin/test_fixtures/github_calls/prs/with_te
 <1> Defines set of regular expressions which will be used to match files changeset and determine if Pull Requests comes with any changed test files.
 <2> These regular expressions will be used against changeset in the Pull Requests to exclude files that don't have to be verified by any test. If only such files exists the check will be marked as "Success" as no tests are expected for such a PR.
 <3> Allows you to decide if you want to combine your patterns with predefined defaults (`true` by default).
+<4> Sets either relative path or absolute URL to a file that contains a message to be added to the Pull Request when no test is found (if nothing is set then a default message is used).
```
bartoszmajsak: How about a link to the default message or even sourcing it here?

pkg/github/raw_file_service.go
```diff
@@ -0,0 +1,24 @@
...
+	Change scm.RepositoryChange
+}
+
+// GetRawFile retrieves raw file content on the given path from the related GitHub repository change
+func (s *RawFileService) GetRawFile(path string) ([]byte, bool, error) {
+	return utils.GetFileFromURL(s.GetRawFileURL(path))
+}
+
+// GetRawFileURL creates a url to the given path related to the GitHub repository change
+func (s *RawFileService) GetRawFileURL(path string) string {
```
bartoszmajsak: Not sure why this should be public?

pkg/github/raw_file_service.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"fmt"
+
+	"github.com/arquillian/ike-prow-plugins/pkg/scm"
+	"github.com/arquillian/ike-prow-plugins/pkg/utils"
+)
+
+// RawFileService encapsulates retrieval of files in the given GitHub repository change
+type RawFileService struct {
+	Change scm.RepositoryChange
```
bartoszmajsak: If we make a `RepositoryChange` as a part of this struct we might want to extend the doc saying it represents the file in the given state of repository indicated by hash.

pkg/plugin/comment_service.go
```diff
@@ -0,0 +1,103 @@
...
+		},
+		commentContext: commentContext,
+	}
+}
+
+// PluginComment checks all present comments in the issue/pull-request. If no comment with IkePluginsTitle is found, then
+// it adds a new comment with IkePluginsTitle, assignee, the plugin tittle and the given commentMsg. If a comment with
+// IkePluginsTitle is found but missing the particular plugin title, then it modifies the comment by adding plugin title
+// with the given commentMsg. If everything is present already, then it does nothing.
+func (s *CommentService) PluginComment(commentMsg string) error {
```
bartoszmajsak: I am not sure if that will work considering all the plugins are notified by the `hook` service at the moment particular event arrives. This might mean we can have a race condition and end up with multiple comments at the same time.

Maybe, for the time being, let's make one comment per plugin to make them independent? We will think about overall comment with references to all active plugins later on. Especially considering the fact that Prow has special `help` website where it lists all plugins (see what is deployed to your openshift cluster)

pkg/plugin/config/config_loader_test.go
```diff
@@ -1,16 +1,18 @@
-package config
+package config_test
 
 import (
+	"github.com/arquillian/ike-prow-plugins/pkg/plugin/config"
 	"github.com/arquillian/ike-prow-plugins/pkg/scm"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	gock "gopkg.in/h2non/gock.v1"
+	"gopkg.in/h2non/gock.v1"
```
bartoszmajsak: Please don't remove that `gock "gopkg.in/h2non/gock.v1"` - as noted in the [docs](http://arquillian.org/ike-prow-plugins/#_building) if we use `goimports` this will fail. Is that IDEA/Goland doing it on your behalf?

pkg/plugin/comment_service.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+			return s.appendToComment(commentMsg, com)
+		}
+	}
+
+	comment := &github.IssueComment{
+		Body: s.append(s.createBeginning(), s.createPluginParagraph(commentMsg)),
+	}
+
+	_, _, err = s.client.Issues.CreateComment(context.Background(), s.issue.Owner, s.issue.RepoName, s.issue.Number, comment)
```
bartoszmajsak: Probably worth logging this error (and all below)?

pkg/utils/url.go
```diff
@@ -0,0 +1,29 @@
...
+	"net/http"
+)
+
+// GetFileFromURL retrieves the content of the file on the given url
+func GetFileFromURL(url string) ([]byte, bool, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, false, err
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
bartoszmajsak: What about `4xx` and `5xx`?

pkg/plugin/test-keeper/plugin/event_handler.go
```diff
@@ -141,33 +131,44 @@ func (gh *GitHubTestEventsHandler) handlePrComment(prComment *gogh.IssueCommentE
...
+
+	cerr := commentService.PluginComment(CreateCommentMessage(url, configuration, change))
+	if cerr != nil {
+		return cerr
 	}
+	return err
+}
 
-	matcher := LoadMatcher(configuration)
+func (gh *GitHubTestEventsHandler) checkTests(change scm.RepositoryChange, config TestKeeperConfiguration, prNumber int) (bool, error) {
```
bartoszmajsak: Can't we have `config` as part of the `GitHubTestEventsHandler` struct?

pkg/plugin/test-keeper/plugin/configuration.go
```diff
@@ -0,0 +1,31 @@
...
+
+	configuration := TestKeeperConfiguration{Combine: true}
+	exists, err := configLoader.Load(&configuration)
+	if err != nil {
+		log.Warnf("Config file was not loaded. Cause: %", err)
+	}
+	if exists {
+		return configLoader.CreateConfigFileURL(), configuration
+	}
+	return "", configuration
```
bartoszmajsak: Why are we returning a tuple here? How about having a new field in the struct itself encapsulating it? Then we would probably don't need `exists` as a returned information too (as I guess `err` should be enough to determine that in general).

pkg/plugin/test-keeper/plugin/comment_message.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"net/url"
+
+	"github.com/arquillian/ike-prow-plugins/pkg/github"
+	"github.com/arquillian/ike-prow-plugins/pkg/scm"
+	"github.com/arquillian/ike-prow-plugins/pkg/utils"
+)
+
+const (
+	beginning = "There wasn't found any test file that would be changed in this PR. Please add a test case to this PR.\n" +
```
bartoszmajsak: You can probably use \` \` for better formatting of those messages. Now it's quite hard to review them as GH doesn't wrap it.

pkg/plugin/comment_service.go
```diff
@@ -0,0 +1,103 @@
...
+type RepositoryIssue struct {
+	Owner    string
+	RepoName string
+	Number   int
+}
+
+// CommentContext holds a plugin name and a assignee to be mentioned in the comment
+type CommentContext struct {
+	PluginName string
+	Assignee   string
```
bartoszmajsak: We might have multiple Assignees in GH (especially reviewers).

pkg/plugin/comment_service.go
```diff
@@ -0,0 +1,103 @@
...
+// CommentService is a struct managing plugin comments
+type CommentService struct {
+	client         *github.Client
+	log            *logrus.Entry
+	issue          RepositoryIssue
+	commentContext CommentContext
+}
+
+// RepositoryIssue holds owner name, repository name and an issue number
+type RepositoryIssue struct {
```
bartoszmajsak: Not sure if that belongs here.

pkg/plugin/comment_service_test.go
```diff
@@ -0,0 +1,134 @@
...
+			service := plugin.NewCommentService(client, test.CreateNullLogger(), change, 2, commentContext)
+
+			// when
+			err := service.PluginComment("New comment")
+
+			// then
+			Ω(err).ShouldNot(HaveOccurred())
+		})
+
+		It("should send patch request with modified comment that contains missing plugin message", func() {
```
bartoszmajsak: Is it important that it's `PATCH` from the feature perspective? It's really an implementation detail ;)

pkg/plugin/comment_service_test.go
```diff
@@ -0,0 +1,134 @@
...
+			}
+			commentContext := plugin.CommentContext{
+				PluginName: "test-keeper",
+				Assignee:   "toAssign",
+			}
+
+			//
+			gock.New("https://api.github.com").
+				Path("/repos/owner/repo/issues/comments/372707978").
+				Reply(400)
```
bartoszmajsak: What does it mean?

pkg/plugin/comment_service_test.go
```diff
@@ -0,0 +1,134 @@
...
+				PluginName: "another-plugin",
+				Assignee:   "toAssign",
+			}
+
+			expContent := "### Ike Plugins\n\n@MatousJobanek Please pay attention to the following message:\n\n" +
+				"#### test-keeper\n\nMessage from test-keeper\n\n#### another-plugin\n\nNew comment"
+
+			toHaveModifiedBody := func(statusPayload map[string]interface{}) bool {
+				return Expect(statusPayload).To(SatisfyAll(
+					test.HaveBody(expContent),
```
bartoszmajsak: Can't we use `.` import for `test` package? We do that in other places so we might want to make it unified.

pkg/plugin/test-keeper/plugin/comment_message.go
bartoszmajsak: I will review those separately and come back with proposal.

pkg/plugin/comment_service.go
```diff
@@ -0,0 +1,102 @@
...
+// PluginComment checks all present comments in the issue/pull-request. If no comment with IkePluginsTitle is found, then
+// it adds a new comment with IkePluginsTitle, assignee, the plugin tittle and the given commentMsg. If a comment with
+// IkePluginsTitle is found but missing the particular plugin title, then it modifies the comment by adding plugin title
+// with the given commentMsg. If everything is present already, then it does nothing.
+func (s *CommentService) PluginComment(commentMsg string) error {
+
+	comments, _, err := s.client.Issues.
+		ListComments(context.Background(), s.issue.Owner, s.issue.RepoName, s.issue.Number, &github.IssueListCommentsOptions{})
+	if err != nil {
+		s.log.Errorf("Getting all comments failed with an error: %s", err)
```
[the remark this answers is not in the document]
MatousJobanek: Not sure - we have two options:
* return the error here and don't try to add the comment, so it is possible that there won't be any comment in the pr
* only log the error and then try to add the comment, so it is possible that there will be more than one error (but at least one)

I chose the second option

docs/test-keeper.adoc
MatousJobanek: I'll probably link them here as the default message is constructed based on the presence of the configuration file, so it would be hard to explain/show it here

pkg/github/raw_file_service.go
MatousJobanek: because I'm using it for constructing URL when the custom-message-file is set as a relative path

pkg/github/raw_file_service.go
MatousJobanek: in the same way as we do it for the config file - make sense, I'll do that

pkg/plugin/comment_service.go
MatousJobanek: I'm aware of the problem. I have an idea how to solve it, but didn't want to try to implement it here. I thought that this could be fine as the first step...

pkg/plugin/config/config_loader_test.go
MatousJobanek: I ran `goimports` before I pushed, so I thought that it would complain about it.

pkg/plugin/comment_service.go
MatousJobanek: isn't it fine to return the err back to the eventHandler?

pkg/utils/url.go
MatousJobanek: correct me if I'm wrong - the check says:
* `resp.StatusCode < 200` - anything that is lower than 200

or
* `resp.StatusCode >= 300` - anything that is higher or equal 300, which means also 4xx and 5xx

then return empty `[]byte` and boolean as false - which means that the request wasn's sucessful

pkg/plugin/test-keeper/plugin/event_handler.go
MatousJobanek: I was thinking about it as well - I wasn't sure if it wouldn't be confusing that when the `GitHubTestEventsHandler` struct is  instantiated, that it would require the config that is not available at that time

pkg/plugin/test-keeper/plugin/configuration.go
MatousJobanek: funny, this was one of my thoughts as well :-D. I decided to use the possibility of returning more than one variables to keep the configuration struct "clean" in the sense that it contains only configurable params... But I'm open to adding it there ;-)

pkg/plugin/comment_service.go
MatousJobanek: I think that I asked you this question - who should be mention in the comment - or maybe I didn't. 
But again, for the purpose of this plugin, this seemed to be sufficient - I didn't want to speculate about the other usages. When we want to bring support for the other plugins, then we can think about more assignees. 
in addition, I'm not sure what that would work - the comment is added once when the PR is opened and the "owner" of the PR is the only person that is connected to it

pkg/plugin/comment_service.go
MatousJobanek: I'm open to any suggestions :-)

pkg/plugin/comment_service_test.go
MatousJobanek: is this better?
`"should modify already existing comment by adding a missing plugin message"`

pkg/plugin/comment_service_test.go
MatousJobanek: That it should not modify the comment - if it tries to modify it, then it should fail.
But it is not necessary maybe, as gock would complain about not-matching request anyway, right?

pkg/plugin/comment_service_test.go
MatousJobanek: sure, I'll do that

pkg/github/raw_file_service.go
MatousJobanek: and for constructing URL to the config file

MatousJobanek: both cases are for the comment purposes

pkg/plugin/config/config_loader_test.go
bartoszmajsak: It might be that it's not stripping the tests. I'm a bit confused with this tool now :)

pkg/plugin/comment_service.go
bartoszmajsak: `// TODO rethink this naming when plugins will start interacting with issue creators and reviewers`

pkg/plugin/comment_service.go
bartoszmajsak: `scm_types.go`

bartoszmajsak:
Action items after peer review:

- [ ] rename comment msg file to `plugin_hint`
- [ ] `GetFileFromUrl` should return `error` in case of `4xx` and `5xx`
- [ ] make `Load` in config to return error
- [ ] one comment per plugin
- [ ] create issue allowing to use `plugin_hint` for inline text and golang template @bartoszmajsak

pkg/plugin/config/config_loader_test.go
MatousJobanek: so, it probably doesn't validate the tests...

pkg/plugin/comment_service_test.go
MatousJobanek: removed as the error with the description of the test case is self-explanatory

pkg/plugin/comment_service.go
MatousJobanek: moved to types.go in scm

pkg/plugin/test-keeper/plugin/comment_message.go
MatousJobanek: using `` won't help much as in some cases I cannot include the new line that this string would keep.
I have split it so it is more readable

pkg/plugin/comment_service.go
MatousJobanek: the error would be logged in the server: https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/plugin/server/server.go#L39
so it is not needed to log it here as well

pkg/github/raw_file_service.go
MatousJobanek: 1956e06

pkg/plugin/comment_service.go
MatousJobanek: comment-per-plugin 797b9b5

pkg/plugin/config/config_loader_test.go
MatousJobanek: d062bb8

pkg/utils/url.go
MatousJobanek: returning error when not matched 525c07a

pkg/plugin/test-keeper/plugin/comment_message.go
MatousJobanek: 934ae04

docs/test-keeper.adoc
MatousJobanek: be96946

pkg/plugin/test-keeper/plugin/configuration.go
MatousJobanek: 3aaf76e

pkg/plugin/comment_service.go
MatousJobanek: e1bed20

pkg/plugin/comment_service.go
MatousJobanek: 938c1f8

pkg/plugin/comment_service_test.go
MatousJobanek: part of comment-per-plugin solution: 797b9b5

pkg/plugin/comment_service_test.go
MatousJobanek: part of comment-per-plugin solution: 797b9b5

pkg/plugin/comment_service_test.go
MatousJobanek: 61c26aa

Issue #43, opened 2018-03-16
Support not only config files with yml extension but also yaml


MatousJobanek:
mainly because `yaml` extensions is the recommended one: http://www.yaml.org/faq.html

PR #44, opened 2018-03-16
feat(#43): adds support for config files with .yaml extension

MatousJobanek:
* Firstly it tries to get file with `yml` extension, then with `yaml`
* introduced `Configuration` interface and default implementation `PluginConfiguration` to enable setting the `LocationURL` directly after the actual download in the `Load` function
* removed some unnecessary function for either creating URL or getting raw files - they are not needed anymore.


MatousJobanek:
> I am not sure about setLocatinUrl idea

there is no other way to set the value when you get the interface

> Why a "java-bean" like interface? Can't we just have it as part of the plugin config structs? 

sure we can - as we had it before, plus setting it as part of the particular plugin config loader. https://github.com/arquillian/ike-prow-plugins/blob/6c037cc3880899911afd6a5f79d22176479907f1/pkg/plugin/test-keeper/plugin/configuration.go#L29 There are two reasons why I moved it to the `Load` method:
* this is something that will be probably common for all plugin configs
* outside of the `Load` method, I don't know which file extension was used - `yml` or `yaml`

bartoszmajsak:
> there is no other way to set the value when you get the interface

I know but so I am not sure about the interface :) somehow it looks a bit strange

MatousJobanek:
> I know but so I am not sure about the interface :) somehow it looks a bit strange

But without the inheritance, I have no idea how to do that.

bartoszmajsak:
I will try to come back with the proposal over the weekend :)

PR #45, opened 2018-03-16
feat(#30): introduces sentry integration

### LINKED ISSUE #30: Add error tracking using Sentry
There is now an error tracking service available for internal OpenShift.io components to consume. Before we go live we should add Sentry integration to capture errors.

For details about the setup and use of Sentry head over to [housekeeping repository](https://gitlab.cee.redhat.com/dtsd/housekeeping/blob/master/docs/errortracking.md).

One place to look at the integration can be https://github.com/fabric8-services/fabric8-wit/pull/1909

bartoszmajsak:
* unifies logging by providing facade interface
  * sets logging context in `serveHTTP` method
  *  `config/sentry.dsn` file should contain sentry DSN, but works without it (errors not sent to sentry)

**IMPORTANT**: move ignored token files to `config/` directory

Fixes #30

### DIFF
--- a/pkg/plugin/pr-sanitizer/main.go
+++ b/pkg/plugin/pr-sanitizer/main.go
@@ -2,10 +2,10 @@ package main
 
 import (
 	gogh "github.com/google/go-github/github"
-	"github.com/sirupsen/logrus"
 	"k8s.io/test-infra/prow/pluginhelp"
 
 	"github.com/arquillian/ike-prow-plugins/pkg/github"
+	"github.com/arquillian/ike-prow-plugins/pkg/log"
 	pluginBootstrap "github.com/arquillian/ike-prow-plugins/pkg/plugin"
 	"github.com/arquillian/ike-prow-plugins/pkg/plugin/server"
 )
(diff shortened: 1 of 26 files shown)


pkg/plugin/pr-sanitizer/main.go
```diff
@@ -2,57 +2,42 @@ package main
...
-	gh.log = logrus.StandardLogger().WithField("ike-plugins", ProwPluginName).WithFields(
-		logrus.Fields{
-			"event-type": eventType,
-			"event-GUID": eventGUID,
-		},
-	)
-
-	gh.log.Info("Handling labels event.")
+func (gh *GitHubLabelsEventsHandler) HandleEvent(log log.Logger, eventType github.EventType, payload []byte) error {
+
```
MatousJobanek: is doing nothing intended?

bartoszmajsak: not sure we need to log it every time we receive event. we have it tested so we are only really interested in errors we should react on I think

bartoszmajsak: ah, ok you mean empty method... well... we haven't started working on this plugin yet :)

pkg/plugin/work-in-progress/plugin/event_handler_unit_test.go
```diff
@@ -1,12 +1,11 @@
 package plugin_test
 
 import (
-	. "github.com/arquillian/ike-prow-plugins/pkg/internal/test"
 	"github.com/google/go-github/github"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
-	gock "gopkg.in/h2non/gock.v1"
+	"gopkg.in/h2non/gock.v1"
```
MatousJobanek: that should not be modified, but as the `goimports` doesn't complain about it then I wouldn't care...

bartoszmajsak: I think it skips the test packages

PR #47, opened 2018-03-19
docs: introduces Sentry section to documentation

bartoszmajsak:
This PR adds a section about Sentry integration to the documentation.

#### Small additions

* locks `raven-go` client
* externalizes logrus and sentry setup to `log` package

### DIFF
--- a/pkg/plugin/plugin_init.go
+++ b/pkg/plugin/plugin_init.go
@@ -14,15 +14,12 @@ import (
 	"github.com/arquillian/ike-prow-plugins/pkg/plugin/server"
 	"github.com/arquillian/ike-prow-plugins/pkg/utils"
 	"github.com/google/go-github/github"
-	"github.com/sirupsen/logrus"
 	"k8s.io/test-infra/prow/pluginhelp/externalplugins"
 	"k8s.io/test-infra/prow/plugins"
 
 	"net/http"
 
-	"time"
-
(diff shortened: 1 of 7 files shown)


pkg/plugin/plugin_init.go
```diff
@@ -55,26 +53,35 @@ func InitPlugin(pluginName string, newEventHandler EventHandlerCreator, newServe
...
+	logger := log.ConfigureLogrus(pluginName)
+	sentryDsn, err := utils.LoadSecret(*sentryDsnSecretFile)
+	if err != nil {
+		logger.WithError(err).Errorf("unable to load sentry dsn from %q. No sentry integration enabled", sentryDsnSecretFile)
+	}
+
+	log.AddSentryHook(logger, log.NewSentryConfiguration(string(sentryDsn), map[string]string{
+		"plugin": pluginName,
+		"environment": *environment,
+	}, *sentryTimeout))
```
MatousJobanek: This whole part could be moved to separated function located either in this file or in the `log` package, no make the `InitPlugin` function shorter.

bartoszmajsak: I thought `log.AddSentryHook` is the function, the rest is just constructing proper config based on init of the plugin (flags).

bartoszmajsak: Eventually I did that as e287906... and ... it had a bug :P

PR #49, opened 2018-03-20
feat(#48): adds repository and sender to logging context

### LINKED ISSUE #48: Extend logging context
As PR automation plugins are intended to run as a service we will be hosting multiple repositories and organizations. Thus for better diagnostic we have to extend logging context to include at least:

- GitHub event GUID (sent with every hook as `X-GitHub-Event` header)
- GitHub event type (sent with every hook as `X-GitHub-Delivery` header)

If applicable (so the event contain this information):

- GitHub repository originating this hook (`repository`)
- User  who triggered this event (`sender`)

bartoszmajsak:
- adds repo and sender to logging context
- unifies status values by using `const`

Fixes #48

### DIFF
--- a/pkg/github/types.go
+++ b/pkg/github/types.go
@@ -10,6 +10,25 @@ type StatusContext struct {
 }
 
 const (
+	// EventGUID is sent by Github in a header of every webhook request.
+	EventGUID = "event-GUID" // aligned with Prow fields
+	// Event is sent by GitHub in a header of every webhook request.
+	Event = "event-type" // aligned with Prow fields
+	// RepoLogField is the repository from where the event came.
+	RepoLogField = "github-repo"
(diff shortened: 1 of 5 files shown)


pkg/github/types.go
```diff
@@ -9,6 +9,25 @@ type StatusContext struct {
 	PluginName string
 }
 
+const (
+	// EventGUID is sent by Github in a header of every webhook request.
+	EventGUID = "event-GUID" // aligned with Prow fields
+	// Event is sent by GitHub in a header of every webhook request.
+	Event = "event-type" // aligned with Prow fields
```
MatousJobanek: Shouldn't that be named `EventType`?

bartoszmajsak: Only if another `EvenType` struct wouldn't be just below ;)

PR #51, opened 2018-03-20
feat(#50): introduces minimal deployment for pilot

### LINKED ISSUE #50: Minimal deployment for pilot
For the pilot deployment, we only need part of the Prow infrastructure to provide automation around PRs and issues through our external plugins.

- `hook` service deployment with route
- plugins

bartoszmajsak:
Tihs PR brings following changes:

- defines hook service template on its own which can be used to deploy on OpenShift
- fixes few minor issues around Sentry
  - lost Makefile bit for registering a secret
  - proper logging of sentry-dsn mount path
- updates documentation with new targets

Fixes #50


Makefile
```diff
@@ -85,25 +85,43 @@ check: ## Concurrently runs a whole bunch of static analysis tools
...
+.PHONY: oc-init-project
+oc-init-project: ## Initiliazes new project with config maps and secrets
 	@echo "Setting cluster project"
-	@oc new-project ike-prow-plugins
+	@oc new-project $(OC_PROJECT_NAME)
 	@oc create configmap plugins && oc create configmap config
-	@oc create secret generic hmac-token --from-file=hmac=config/hmac.token && oc create secret generic oauth-token --from-file=oauth=config/oauth.token
+	@oc create secret generic hmac-token --from-file=hmac=config/hmac.token
+	@oc create secret generic oauth-token --from-file=oauth=config/oauth.token
+	@oc create secret generic sentry-dsn --from-file=sentry=config/sentry.dsn
```
MatousJobanek: in the README, there is no mention about the `sentry.dsn` file. It is only in the documentation: http://arquillian.org/ike-prow-plugins/#_sentry_logging where is written: `Plugins in this repository can be integrated with Sentry` which means that Sentry should be optional. 
When I run it locally on minishift, I guess that Sentry is not required, right? At least it is not mentioned in the prerequisites. But the Makefile fails when it tries to load the `sentry.dsn` file.

bartoszmajsak: Yeah, noticed that. Will think how to fix it and also cleanup secret creation as it's very much copy-paste

bartoszmajsak: Created new issue #52

bartoszmajsak: Brought `Makefile` improvements in https://github.com/arquillian/ike-prow-plugins/pull/53

bartoszmajsak:
@MatousJobanek please check if it now works on your machine and documentation has all important bits in.

plugins.yaml
```diff
@@ -2,12 +2,14 @@ plugins:
   bartoszmajsak/wfswarm-booster-pipeline-test:
   - size
 
+# tag::externalPlugins[]
 external_plugins:
   bartoszmajsak/wfswarm-booster-pipeline-test:
```
MatousJobanek: when this file is included in the README, then it would be good to change it to something general: 
`organization/repository`

bartoszmajsak: The reason why I include this file is to have it also testable, for that purpose it references to an actual existing repository. We can change it in the future, but I would still prefer to have it as "living" config. Otherwise, we can simply just revert it to be just a text in the doc.

README.adoc
```diff
@@ -129,18 +96,18 @@ by `oc whoami -t` and is used as password by Docker daemon. To do that open Open
...
-
-
 NOTE: If you are using `zsh` you can take a look at link:https://github.com/kennethreitz/autoenv[autoenv] or
 link:https://direnv.net/[direnv] to automate this setup every time you access project's directory. Just make sure to add it to your global `.gitignore` to avoid sharing it through PRs (e.g. `git config --global core.excludesfile ~/.gitignore`).
 
+[#sample-autoenv]
+[source,bash]
+.Sample shell script with local dev setup
+----
+include::local.sh[]
```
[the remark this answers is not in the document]
bartoszmajsak: It seems that GH replaces it to `link:` on the fly without this being an actual link.

plugins.yaml
MatousJobanek: make sense, keep it as it is

README.adoc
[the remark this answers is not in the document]
bartoszmajsak: And yeah, `include::` macro is blocked by GH `.adoc` rendering due to security reasons I could assume.

plugins.yaml
bartoszmajsak: But yeah, don't want to have it as my name everywhere. Let's change this and tests soon.

PR #56, opened 2018-03-21
fix(#54): status improvements

### LINKED ISSUE #54: Make status messages more contextual
Currently, we have the same status message used for multiple cases which can be confusing at times.
For example when PR contains only "legit files", like `pom.xml` for which you most likely don't ship tests, it should have different msg than "There are some tests :)".

Additional improvements:

  * status context shouldn't contain quoted context (now it's `"ike-plugins"/"test-keeper" — There are some tests :)`)

bartoszmajsak:
- makes status context without extra quotes
- counts excluded files to treat it as different status message
- renames bot to alien-ike

Fixes #54


pkg/plugin/test-keeper/plugin/event_handler_gh_test.go
```diff
@@ -235,7 +230,7 @@ var _ = Describe("Test Keeper Plugin features", func() {
 
 			gock.New("https://api.github.com").
 				Post("/repos/" + repositoryName + "/statuses").
-				SetMatcher(ExpectPayload(toHaveFailureState)).
+				SetMatcher(ExpectPayload(toBe(github.StatusFailure, "No tests in this PR :("))).
```
MatousJobanek: would make sense to use constants for these strings?

pkg/plugin/test-keeper/plugin/test_checker.go
```diff
@@ -12,23 +12,48 @@ type TestChecker struct {
...
-func (checker *TestChecker) IsAnyNotExcludedFileTest(files []scm.ChangedFile) (bool, error) {
+// FileCategories holds information about the total files coming in the changeset, legit files (those which are excluded from test verification)
+// and Tests
+type FileCategories struct {
+	Total, Legit, Tests int
+	Files               *[]scm.ChangedFile
+}
+
+// OnlyLegitFiles indicates if changeset contains only files which are excluded from test verification
+func (f *FileCategories) OnlyLegitFiles() bool {
```
MatousJobanek: aren't they more likely excluded or skipped files? Legit is (for me) something very vague and general.

pkg/plugin/test-keeper/plugin/test_checker.go
```diff
@@ -12,23 +12,48 @@ type TestChecker struct {
...
+	return f.Total > 0 && f.Legit == f.Total
```
MatousJobanek: do we really need to check the total count?
btw, if there is a PR without any file different, then it is automatically merged/closed

pkg/plugin/test-keeper/plugin/test_checker.go
```diff
@@ -12,23 +12,48 @@ type TestChecker struct {
...
+
+// NewFileCategories creates new instance of FileCategories struct with files populated
+func NewFileCategories(files []scm.ChangedFile) FileCategories {
+	return FileCategories{Files: &files, Total: len(files)}
+}
 
-	remainingNoTestFiles := false
+// CategorizeFiles counts files in the changeset which are tests (included files) and should not be considered for
+// verification (excluded). When first test is found it stops, as this is enough to unblock PR
+func (checker *TestChecker) CategorizeFiles(files []scm.ChangedFile) (FileCategories, error) {
```
MatousJobanek: Is this still `TestChecker` or something else? 
btw, when I read `CategorizeFiles` I would expect that I get some categories - slices of files separated into different categories, not only count of the file types. Would make sense to rename it?:
`TestChecker` -> `FileTypeCounter`
`CategorizeFiles` -> `Count`
`FileCategories` -> `FileTypes` or `CountedFileTypes`

I know that I'm not good at naming stuff, my proposal is maybe worse than yours, it just didn't make sense to me at the very first look.

pkg/plugin/test-keeper/plugin/event_handler_gh_test.go
bartoszmajsak: Sure, good idea. On the way

pkg/plugin/test-keeper/plugin/test_checker.go
bartoszmajsak: Skipped is a good name! Thanks

pkg/plugin/test-keeper/plugin/test_checker.go
bartoszmajsak: There won't be 0 files in GitHub PR, that is true. But for the boundary case, if the changeset has 0 files than it have no tests and no skipped files - so answering to "onlySkippedFiles" should be false, right?

pkg/plugin/test-keeper/plugin/test_checker.go
bartoszmajsak: That's right, maybe `Categorize` is too much for just counting them up. Let me reword it.

pkg/plugin/test-keeper/plugin/test_checker.go
MatousJobanek: Agree. 
Last question  - do we need to also check the boundary case when there is no changed file in the PR - in the event_handler? Using common sense I would say no. As a pedantic guy probably yes :-D

pkg/plugin/test-keeper/plugin/event_handler_gh_test.go
bartoszmajsak: def5e6c

bartoszmajsak: def5e6c

pkg/plugin/test-keeper/plugin/test_checker.go
bartoszmajsak: e5df495

pkg/plugin/test-keeper/plugin/test_checker.go
bartoszmajsak: That's unrealistic assuming it is coming from GitHub... So I would say YAGNI? ;)

pkg/plugin/test-keeper/plugin/test_checker.go
bartoszmajsak: e9377ce

PR #58, opened 2018-03-22
chore: test improvements

bartoszmajsak:
This PR brings improvement for data-driven tests by simplifying the creation of test cases and extending check for default inclusions.

In addition `node_modules` has been added as excluded build dir and thus test case checking if PR is not accepted only if tests in changset are in the "vendor" directories.


bartoszmajsak:
@MatousJobanek aad4a37 might be a bit over the top :) so let's consider aad4a37 vs c599a4b, where aad is with the aim of more natural reading

pkg/plugin/test-keeper/plugin/matchers_test.go
```diff
@@ -40,174 +101,82 @@ var _ = Describe("Test Matcher features", func() {
...
-			createEntry(DefaultMatchers.Exclusion, "src/github.com/arquillian/ike-prow-plugins/Makefile", true),
-			createEntry(DefaultMatchers.Exclusion, "src/main/java/pom.xml", true),
-			createEntry(DefaultMatchers.Exclusion, "mvnw", true),
-			createEntry(DefaultMatchers.Exclusion, "mvnw.cmd", true),
-			createEntry(DefaultMatchers.Exclusion, "mvnw.bat", true),
-			createEntry(DefaultMatchers.Exclusion, "build.gradle", true),
-			createEntry(DefaultMatchers.Exclusion, "gulpfile.js", true),
+		table.DescribeTable("should exclude common build tools",
+			expectThatFile,
+			matches(DefaultMatchers.Exclusion).using(buildAssets)...,
```
bartoszmajsak: Another way could be

```
expectThatFile,
from(buildAssets).matches(DefaultMatchers.Exclusion)....,
```

MatousJobanek:
> aad4a37 might be a bit over the top :) so let's consider aad4a37 vs c599a4b, where aad is with the aim of more natural reading

I would keep it as it is - I like it
Good job

PR #60, opened 2018-03-23
fix(#23): adds html, shell and docker patterns

### LINKED ISSUE #23: Exclude PRs from test check if they only contain defined type of files
We should be able to define set of file extensions which, if only they are present in the PR changes, should be omitted by the plugin scan. In particular, this can be useful for PRs containing build setup update, documentation, styling etc.

Extend `test-keeper.yaml` configuration to allow such exception. The user should be able to define a list of simplified regular expressions which can be applied on files. For example:

```yaml
excludes: *.adoc,*.md,*.css
```

bartoszmajsak:
Fixes #23

### DIFF
--- a/pkg/plugin/test-keeper/plugin/matchers.go
+++ b/pkg/plugin/test-keeper/plugin/matchers.go
@@ -48,9 +48,9 @@ func Matches(matchers []FileNamePattern, filename string) bool {
 var DefaultMatchers = TestMatcher{
 	Inclusion: []FileNamePattern{javaTests, goTests, javascriptTests, typescriptTests, pythonTests, groovyTests},
 	Exclusion: []FileNamePattern{
-		buildToolsFileNameMatcher, buildToolsDirectoryNameMatcher,
-		ciToolsFileNameMatcher, settingsFileNameMatcher,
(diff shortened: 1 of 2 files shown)


pkg/plugin/test-keeper/plugin/matchers.go
```diff
@@ -78,26 +78,30 @@ var groovyTests = FileNamePattern{
...
 
-var ciToolsFileNameMatcher = FileNamePattern{
+var ciTools = FileNamePattern{
 	Regex: `\.travis\.yml|Jenkinsfile|\.gitlab-ci\.yml,|wercker\.yml|circle\.yml$`,
 }
 
-var uiAssetsFileNameMatcher = FileNamePattern{
-	Regex: `(\.jpg|\.jpeg|\.png|\.ico|\.svg|\.gif|\.css|\.scss|\.sass|\.less)$`,
+var uiAssets = FileNamePattern{
+	Regex: `(\.jpg|\.jpeg|\.png|\.ico|\.svg|\.gif|\.css|\.scss|\.sass|\.less|\.[xsd]?html|\.ejs)$`,
```
bartoszmajsak: `shtml` - server side includes html
`ejs` - js templating

PR #63, opened 2018-03-23
feat: adds Ike reaction to "no tests" comment

bartoszmajsak:
plus custom string `StartsWith` gomega assertion for tests.


pkg/plugin/test-keeper/plugin/comment_message.go
```diff
@@ -29,18 +29,22 @@ const (
 		"but the plugin wasn't able to retrieve it from the defined location (%s). Make sure it is either a valid URL or a " +
 		"path to an existing file in this repository."
 
+	sadIke = `<p align="center">
+  				<img src="http://design.jboss.org/arquillian/logo/ui/images/failure/arquillian_ui_failure_256px.svg">
+			  </p>`
```
MatousJobanek: the image is too big to my taste - it doesn't bring any additional value to the comment, but occupy 50% of it. I'm not against to having it there, but I would make it either smaller or make the text flowed around the image:
obrazek
the code is ``

pkg/internal/test/starts_with_matcher.go
```diff
@@ -0,0 +1,54 @@
...
+	prefix string
+}
+
+func (matcher *startsWith) Match(actual interface{}) (success bool, err error) {
+	actualString, ok := toString(actual)
+	if !ok {
+		return false, fmt.Errorf("StartsWith matcher requires a string or stringer.  Got:\n%s", format.Object(actual, 1))
+	}
+
+	return strings.Contains(actualString, matcher.prefix), nil
```
MatousJobanek: shouldn't that be `strings.HasPrefix` instead of the `Contains` when the matcher is named `StartWith`?

pkg/internal/test/starts_with_matcher.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/onsi/gomega/format"
+	"github.com/onsi/gomega/types"
+)
+
+// StartWith is a GomegaMatcher verifying if given string starts with prefix
+func StartWith(expected string) types.GomegaMatcher {
```
MatousJobanek: How about `StartsWith` in the 3rd person form - wouldn't that sound more natural?

pkg/internal/test/starts_with_matcher.go
bartoszmajsak: Good catch, not sure why it's not `HasPrefix`

pkg/plugin/test-keeper/plugin/comment_message.go
bartoszmajsak: Sure, let's make it smaller

pkg/internal/test/starts_with_matcher.go
bartoszmajsak: I asked myself the very same question, but that doesn't sound proper English to me, "Expect something to start with" is what we should be reading here.

pkg/internal/test/starts_with_matcher.go
bartoszmajsak: Fixed with 09dadcc

PR #64, opened 2018-03-25
chore(#36): moves config and server to root pkg

### LINKED ISSUE #36: Flatten plugin/ directories
Each plugin has `plugin/` subdirectory where the core logic lives. It's inherited from the original prow code base and it's not really needed. The goal of this task would be to remove this one layer of directory hierarchy

bartoszmajsak:
Moving all the files to live next to `main.go` will end up having them in `main` package and thus we won't be able to import them in other plugins (if ever needed - not sure about that yet). I think for the moment layout for every plugin such as:

```
├── main.go
└── plugin
    ├── event_handler.go
    └──.....
```

should be fine.

Fixes #36


pkg/config/config_suite_test.go
```diff
@@ -1,4 +1,4 @@
-package config_test
+package config
```
MatousJobanek: This shouldn't have been changed I guess

MatousJobanek:
> Moving all the files to live next to main.go will end up having them in main package 

How about doing it the other way around - to move all the files into the plugin's main directory and the `main.go` file into a subdirectory  ?
```
pkg/
├── plugin
│   ├── test-keeper
│   │   ├── comment_message.go
│   │   ├── comment_message_test.go
│   │   ├── configuration_gh_test.go
│   │   ├── configuration.go
│   │   ├── event_handler_gh_test.go
│   │   ├── event_handler.go
│   │   ├── file_categories.go
│   │   ├── file_categories_test.go
│   │   ├── main
│   │   │   └── main.go
│   │   ├── matchers.go
│   │   ├── matchers_test.go
│   │   ├── test_fixtures
│   │   │   └── github_calls
............
│   │   ├── test_keeper_suite_test.go
│   │   └── test_status_service.go
```

bartoszmajsak:
I am not sure if that brings more clarity, to be honest. I would prefer current approach

pkg/config/config_suite_test.go
bartoszmajsak: Right, I think GoLand is messing those up constantly.

bartoszmajsak: Fixed ae5fb65

MatousJobanek:
> Did you try to change it this way? Can you build it?

yes, I tried it for the `test-keeper` and the compilation was OK. 

> I am not sure if that brings more clarity, to be honest.

the idea was to bring more frequently used files closer to the root directory. As for the clarity - you are right that I just swapped the files and the depth of the directory structure stayed same. But I got rid of the duplication of the "plugin" word in the dir path:
`./pkg/plugin/test-keeper/plugin/...`

> I would prefer current approach.

It was just a proposal - I don't insist on it, I'm fine also with the current state

bartoszmajsak:
> It was just a proposal - I don't insist on it, I'm fine also with the current state

I think we eventually gonna find the right structure :)

Issue #21, opened 2018-03-05
Configure junit reporting for Ginkgo tests


bartoszmajsak:
Ginkgo can report tests execution as `junit.xml`. This format can be understood by reporting tools of e.g. CI servers. Configure ginkgo suites to produce junit reports as build artifacts. 

See [Ginkgo documentation](https://onsi.github.io/ginkgo/#generating-junit-xml-output) for more details.

PR #65, opened 2018-03-26
feat(#21): configures JUnit reporting for Ginkgo tests.

hemanik:
Configures JUnit reporting for Ginkgo Tests to produce JUnit reports as build artifacts in a format understood by CI servers. 

Fixes: #21

### DIFF
--- a/pkg/plugin/config/config_suite_test.go
+++ b/pkg/plugin/config/config_suite_test.go
@@ -4,10 +4,12 @@ import (
 	"testing"
 
 	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/reporters"
 	. "github.com/onsi/gomega"
 )
 
 func TestConfig(t *testing.T) {
 	RegisterFailHandler(Fail)
-	RunSpecs(t, "Plugin Config Loader Suite")
+	junitReporter := reporters.NewJUnitReporter("junit.xml")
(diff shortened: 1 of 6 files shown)


bartoszmajsak:
Thanks @hemanik for this improvement. I was just wondering if we couldn't name `junit.xml` better. This name is a bit confusing as it doesn't reveal what the file is about but also hints that there is some JUnit involved :) (we know it's b/c of the xml format, but that's very much hindden).

That said - how about `ginkgo-test-results.xml` instead?

pkg/plugin/config/config_suite_test.go
```diff
@@ -4,10 +4,12 @@ import (
...
 	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/reporters"
 	. "github.com/onsi/gomega"
 )
 
 func TestConfig(t *testing.T) {
 	RegisterFailHandler(Fail)
-	RunSpecs(t, "Plugin Config Loader Suite")
+	junitReporter := reporters.NewJUnitReporter("junit.xml")
+	RunSpecsWithDefaultAndCustomReporters(t, "Plugin Config Loader Suite", []Reporter{junitReporter})
```
bartoszmajsak: Can we move it to `internal/test` package and re-use across all the suites instead?

PR #66, opened 2018-03-26
feat(#35): introduces wildcard patterns as another version of file patterns

### LINKED ISSUE #35: Regex simplification
We should provide simplification for test patterns and excluded files in the similar fashion as [Maven Surefire](http://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html) does.

All types defined below could be combined as a list, separated by commas.

#### Support for wildcards

Wildcards (glob patterns) are the easiest expressions and users are most likely familiar with them. Using it will let them define filename and directory patterns, for example:

  * "**/Test*.java" - includes all of its subdirectories and all Java filenames that start with "Test".
  * "**/*Test.java" - includes all of its subdirectories and all Java filenames that end with "Test".
  * "**/*Tests.java" - includes all of its subdirectories and all Java filenames that end with "Tests".
  *  "**/*TestCase.java" - includes all of its subdirectories and all Java filenames that end with "TestCase".

#### Regular expressions

If a user would like to use standard regular expressions, it should be surrounded by 
`regex{{}}`, e.g. `regex{{.*(Test|IT).java$}}`

#### Negations

Prepending a pattern (in whatever form) with `!` should result in negating the expression (so having opposite effect.

MatousJobanek:
This PR brings following improvements

* Added a possibility to specify file patterns (for inclusions or exclusions) in a wildcard format. If the wildcard format is used, then the pattern is transformed to a regex and used as a matcher.
* inclusions and exclusions parameters are configured as a list of strings
* moved default inclusion and exclusion patterns to a separated file used as a default config file

Fixes #35


pkg/plugin/test-keeper/plugin/file_pattern.go
```diff
@@ -0,0 +1,92 @@
...
+	anyNameRegex          = "[^/]*"
+	anythingRegex         = ".*"
+	twoStarsReplacement   = "<two-stars-replacement>"
+	endOfLineRegex        = "$"
+	directorySeparator    = "/"
+)
+
+// FilePattern contains regex that matches a file
+type FilePattern struct {
+	Regex string
```
bartoszmajsak: Minor thing, but if in go we call it `regexp` then maybe worth to use the same across?

pkg/plugin/test-keeper/plugin/file_pattern_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	Context("File pattern parsing", func() {
+
+		table.DescribeTable(
+			"should parse file patterns to regex",
+			it,
+			should().
+				parsePattern("**/*Test.java").
+				toRegex(`.*/[^/]*Test\.java$`),
```
bartoszmajsak: I'm a bit unsure about this syntax anymore, as it contradicts with the actual types. `it`  is, in fact, an assertion. But one can say a similar thing about `matchers_test.go`. Thus let's treat this comment as "Call To Action" to rethink this and make it best of both words - nice readability but also reflecting what is really going on under the hood

pkg/plugin/test-keeper/plugin/file_pattern_test.go
```diff
@@ -0,0 +1,76 @@
...
+var _ = Describe("File pattern features", func() {
+
+	var it = func(filePattern, expectedRegex string) {
+		parsed := plugin.ParseFilePatterns([]string{filePattern})
+		Expect(parsed).To(ConsistOf(plugin.FilePattern{Regex: expectedRegex}))
+	}
+
+	Context("File pattern parsing", func() {
+
+		table.DescribeTable(
```
bartoszmajsak: Can we also think about covering `func (matcher *FilePattern) Matches(filename string) bool` the same way?

pkg/config/config_loader.go
```diff
@@ -31,3 +33,19 @@ func Load(target interface{}, loader SourcesProvider) error {
...
+
+// IkeProwLocalLoadableConfig holds absolute path to a local config file (located in ike-prow-plugins project) to be loaded
+type IkeProwLocalLoadableConfig struct {
+	AbsFilePath string
+}
+
+// Sources loads local config file that is located in ike-prow-plugins project structure
+func (i *IkeProwLocalLoadableConfig) Sources() []Source {
+	return []Source{func() ([]byte, error) {
+		file, err := ioutil.ReadFile(i.AbsFilePath)
```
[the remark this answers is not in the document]
MatousJobanek: > I'm afraid this will not work when we run just the binary (or ship it as docker image). We have to make this static resource available somehow.

Good point - I didn't realize that...

> We have two options I can think of

there is a third option, but probably not the best - to get the file from GH.

> I would personally opt for the second option. I'm aware suggested library not really maintained anymore, but other OSIO services are using it without any issues so far. Fairly simple intro on how to set it up can be found here: http://rachbelaid.com/embedding-assets-in-go-project/

if it is used in other OSIO projects without any problem, then I'm fine with it

> I could think of creating a common place for static content for all the plugins. 

one common place was an option I was thinking about as well. 

> We can then also use it for templates. How about the following structure?

what kind of templates? Templates that are in the cluster directory or something else? I'm asking as I would keep them where they are - it makes more sense to me.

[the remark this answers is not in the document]
bartoszmajsak: Templates for messages

pkg/plugin/test-keeper/plugin/file_pattern.go
MatousJobanek: but is `regexp` more likely a name of the package/object used in the library or a convention how to name regex expressions...?

pkg/plugin/test-keeper/plugin/file_pattern_test.go
MatousJobanek: > it is, in fact, an assertion. But one can say a similar thing about matchers_test.go

Yes, I have the very same feeling when I saw your code in `matchers_test.go`. this is also why I went with this approach.
How about this:
```go
table.DescribeTable(
	"should parse file patterns to regex",
	assertThat,
	pattern("**/*Test.java").isParsedToRegex(`.*/[^/]*Test\.java$`),
	pattern("*/*Test.java").isParsedToRegex(`[^/]*/[^/]*Test\.java$`),
```

pkg/plugin/test-keeper/plugin/file_pattern_test.go
MatousJobanek: I was thinking about it. I had also an idea to somehow incorporate it into the very first assertion - something like:
```go
it,
should().
	parsePattern("**/*Test.java").
	toRegex(`.*/[^/]*Test\.java$`, 
		matching(
			"src/test/java/org/MyTest.java",
			"!src/test/java/org/AnyClass.java"),
```
but 

* I didn't want to mix two things in one test.
* I realized that the logic of `Matches` method itself is verified in `matchers_test.go` anyway
* I had problems with writing the test that would bring an additional quality but didn't introduce too many code duplication and a lot work (I know I'm lazy)

so I decided to change a few file patterns in several other tests, so the logic would be verified.

pkg/plugin/test-keeper/plugin/file_pattern_test.go
bartoszmajsak: Yes, this is way better

pkg/plugin/test-keeper/plugin/file_pattern_test.go
bartoszmajsak: Yes, we cover it level higher, so that's good. Maybe we can review existing tests and see if we are not missing some important cases.

Issue #75, opened 2018-03-29
Handle occasional 404 returned from GH


MatousJobanek:
sometimes GH returns 404 when plugins are getting or sending some information. We should handle such corner cases in the sense that when the status code is >=404  then the request should be resent - there could be three attempts of resending the requests, if none of them is successful, then it could fail.

PR #76, opened 2018-03-29
feat(#75): resends GH requests when response code is >= 404

MatousJobanek:
* introduced `Client` in `github` package that invokes real GH client
* if response returned from GH has status code >= 404, then it resends the request again - maximal number of attempts is 3
* this new client is used in all plugins instead of the original one because:
  * we can alter everything at once - the number of retries and also which requests is the resend logic applied on - currently on all of them
  * it will make easier any additional types of scm clients introduced in future
  * it decreases the number of parameters

fixes: #75

### DIFF
--- a/pkg/github/client.go
+++ b/pkg/github/client.go
@@ -6,21 +6,29 @@ import (
 	"github.com/arquillian/ike-prow-plugins/pkg/scm"
 	gogh "github.com/google/go-github/github"
 	"golang.org/x/oauth2"
+	"time"
+	"github.com/arquillian/ike-prow-plugins/pkg/http"
 )
 
 // Client manages communication with the GitHub API.
 type Client struct {
-	Client *gogh.Client
+	Client  *gogh.Client
+	Retries int
+	Sleep   time.Duration
 }
 
-// NewClient creates a Client instance with the given oauth secret used as a access token
-func NewClient(oauthSecret []byte) *Client {
+// NewClient creates a Client instance with the given oauth secret used as a access token.
+// The given number of retries and a duration of sleep sets how many times the client should invoke the request until
+// it gets response with a code < 404 and how long it should sleep between every request attempt
+func NewClient(oauthSecret []byte, retries int, sleep time.Duration) *Client {
 	token := oauth2.StaticTokenSource(
 		&oauth2.Token{AccessToken: string(oauthSecret)},
 	)
 
 	return &Client{
-		Client: gogh.NewClient(oauth2.NewClient(context.Background(), token)),
+		Client:  gogh.NewClient(oauth2.NewClient(context.Background(), token)),
+		Retries: retries,
+		Sleep:   sleep,
 	}
 }
 
@@ -29,7 +37,7 @@ func (c *Client) GetPermissionLevel(owner, repo, user string) (*gogh.RepositoryP
 	var permissionLevel *gogh.RepositoryPermissionLevel
 	var err error
 
-	invokeThreeTimesIfNotSuccess(func() *gogh.Response {
+	c.do(func() *gogh.Response {
 		var response *gogh.Response
 		permissionLevel, response, err = c.Client.Repositories.GetPermissionLevel(context.Background(), owner, repo, user)
 		return response
@@ -43,7 +51,7 @@ func (c *Client) GetPullRequest(owner, repo string, prNumber int) (*gogh.PullReq
 	var pr *gogh.PullRequest
 	var err error
 
-	invokeThreeTimesIfNotSuccess(func() *gogh.Response {
+	c.do(func() *gogh.Response {
 		var response *gogh.Response
 		pr, response, err = c.Client.PullRequests.Get(context.Background(), owner, repo, prNumber)
 		return response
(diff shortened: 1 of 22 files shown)


pkg/internal/test/test_doubles.go
```diff
@@ -50,6 +52,13 @@ func ExpectPayload(payloadAssert func(payload map[string]interface{}) bool) gock
 	return matcher
 }
 
+//nolint
+func CreateEmptyGitHubClient() *github.Client {
```
bartoszmajsak: Maybe worth changing `nolint` to something helpful here? Mentioning what this "empty" really means. Maybe also worth to rethink the naming here for two reasons:

- empty doesnt really say what it does? Is it not calling real GH so it's `noop`/`dummy` implementation?
- how about following the Go convention, it's a constructor function, so `New....Client()` would fit.

I will do the same for the logger below in the separated PR (need some new gh karma :D).

pkg/github/client.go
```diff
@@ -0,0 +1,126 @@
...
+			c.Client.Repositories.CreateStatus(context.Background(), change.Owner, change.RepoName, change.Hash, repoStatus)
+		return response
+	})
+
+	return err
+}
+
+type requestInvoker func() *gogh.Response
+
+func invokeThreeTimesIfNotSuccess(invokeRequest requestInvoker) {
```
bartoszmajsak: Since we are on re-tries I would think about making it general purpose.

```golang
// amount of retries, sleep before another one, function to call
retry.Do(3, 1 * time.Second, func() (err error) {
  // actual call
}
```

and then move it away to its own package or maybe re-use existing library (even though it sounds like a rather easy piece of code to write).

WDYT?

pkg/plugin/work-in-progress/plugin/event_handler_unit_test.go
```diff
@@ -1,8 +1,8 @@
 package plugin_test
 
 import (
+	"github.com/arquillian/ike-prow-plugins/pkg/internal/test"
```
bartoszmajsak: Minor thing, in `test-keeper/event_handler_gh_test.go` et al  we `.` import it - I think it would be good going one way or the other and in fact I think `test.` sounds a bit better. Can you align it?

pkg/github/client.go
bartoszmajsak: Then retries and sleep can be part of the `Client` struct and we could make it configurable ease in the `plugin_init` later if needed.

bartoszmajsak: We might also want to log if we fail after all retries.

pkg/internal/test/test_doubles.go
bartoszmajsak: https://github.com/arquillian/ike-prow-plugins/pull/79

pkg/github/client.go
MatousJobanek: > Since we are on re-tries I would think about making it general purpose. and then move it away to its own package

good idea, I'll move it to `util` directory and set the package to `retry`, would that work?

What do you think that should be the default timeout for GH client? 1 sec?

> re-use existing library (even though it sounds like a rather easy piece of code to write).

I took a look at some libraries, but those are focused (developed) for the purpose of using it directly with the http requests - our use-case is slightly different - we don't provide either url nor the request, but a function that returns a response. I didn't spend much time with looking for such a library (or piece of an existing code) as it seemed to be very easy, so coding it took me less time than looking for the libraries

> Then retries and sleep can be part of the Client struct and we could make it configurable ease in the plugin_init later if needed.

+1

> We might also want to log if we fail after all retries.

I was thinking about it, but as we log all the errors (if necessary) in the callers, then it didn't seem to be important to log it here as well. However, log it with the `DEBUG` level could be fine.

MatousJobanek:
> 404 can be related to permission issue (as it seems GH uses that status code to not "leak" permission model based on my observations)

you get 404 only when you want to access a private repo, otherwise, you get 401 or 403 - those are also handled by go-github library.

> Could be good to resolve it in Sentry and link to the issue / PR once merged

Resolve it once PR is merged and a new version of plugins also deployed I guess. I'll link the PR/issue there.

pkg/github/client.go
bartoszmajsak: > What do you think that should be the default timeout for GH client? 1 sec?

Yeah, that is probably good starting point

> I took a look at some libraries

I look at some too [[1](https://github.com/avast/retry-go)][[2](https://medium.com/@matryer/retrying-in-golang-quicktip-f688d00e650a)], but I would still consider writing this on our own.

> I was thinking about it, but as we log all the errors (if necessary) in the callers, then it didn't seem to be important to log it here as well. However, log it with the `DEBUG` level could be fine.

Except that there is no `DEBUG` level ;) You are right that we log errors, but now if we reach the limit of retries what kind of error do we return to the caller? We should make sure we propagate the right one.

pkg/internal/test/test_doubles.go
MatousJobanek: > empty doesnt really say what it does? Is it not calling real GH so it's noop/dummy implementation?

Yes, empty was a really stupid name. 
Technically speaking, it is just a default GH client (without any token set) and it calls real GH, just without any authentication. So I'll call the function `NewDefaultGitHubClient`

MatousJobanek: done 8baf6d3

pkg/github/client.go
MatousJobanek: first two points implemented here: 257d959
it is not possible to have multiple packages inside of one directory, so I created a new dir `http`, put the `retry.go` file there and also moved `url.go` file into this dir.
Now, I'm going to implement returning of the error

pkg/plugin/work-in-progress/plugin/event_handler_unit_test.go
MatousJobanek: done aae7a75

Issue #72, opened 2018-03-27
Explain each state of the test-keeper plugin


bartoszmajsak:
We should provide better insights on why PR was approved or rejected by Test Keeper plugin.

Status API let us add "details" (through `target_url`). We can use it to link to new sections in the [documentation](http://arquillian.org/ike-prow-plugins/#_test_keeper_plugin). I would think about "FAQ" with two questions (and anchor links to them):

-"Why my PR was rejected" section in case of failure explaining how do we verify test presence
- Short section in case of Success.

bartoszmajsak:
I think we should have dedicated page for each state with only this information - pointing to the doc section won't give you this information at hand - there's quite a bit of text around and the relevant piece might not be on the top of the page (depending on the resolution) and it's also not highlighted.

PR #77, opened 2018-03-30
feat(#72): adds details to each state of ike plugins

MatousJobanek:
#### Changes introduced in this PR
* adds "TargetURL" to the repo status that is sent when the evaluation is finished
* applied to all state types except for the pending status
* added a part to the documentation called "Status details" with subsections describing all types of the plugin's state - each of them is referenced by an anchor
* in case of "failure" state, the general link to the Sentry is set
  * general because I don't know how to reference the particular event and I don't know how to figure it out without polluting the Sentry log with an unnecessary set of testing events. And also because I think that this could be part of another story.
  * the link is parsed/created when the `sentry.dsn` file is loaded
* added documentation for work-in-progress plugin to provide the explanation for this plugin as well.

#### What I did differently than it was described in the issue
* Name of the section in the doc - FAQ just seemed to be too generic and focused on something else - eg.: struggling with the setup. In addition, I would expect that FAQ would be one section for the whole project, not for each plugin separated
* Added more subsections than just two to give more focused information for the particular state
* Added the support also for work-in-progress plugin  as the plugin is using the same `StatusService` so it was easier to do it for both plugins at once 

Note: feel free to change the wording in any of the messages without asking me ;-)

Fixes: #72

### DIFF
--- a/docs/fragments/work-in-progress.adoc
+++ b/docs/fragments/work-in-progress.adoc
@@ -1,0 +1,17 @@
+== Work In Progress plugin
+
+This plugin can help reviewers with focusing on work that really matters so they don't need to spend time with Pull Requests that are still in progress.
+
+If the PR title starts with "const:pkg/plugin/work-in-progress/plugin/event_handler.go[name="WipPrefix"]" prefix (non-case sensitive) then the plugin marks its check as **Failure**, or as **Success** otherwise.
+
+=== Status details
+
+==== Success [[wip-success]]
+
+Your Pull Request has been approved as the plugin didn't detect any sign of work in progress (the PR title doesn't start with "const:pkg/plugin/work-in-progress/plugin/event_handler.go[name="WipPrefix"]" prefix)
+
+==== Failed [[wip-failed]]
+
+Your Pull Request has been rejected because the plugin detected that the PR title starts with "const:pkg/plugin/work-in-progress/plugin/event_handler.go[name="WipPrefix"]" prefix, so it seems that there is still an ongoing work on this PR.
+
+When the PR is done and ready for review and merge, then (to make the check status green) remove the prefix from the title.
--- a/pkg/github/status_service.go
+++ b/pkg/github/status_service.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/arquillian/ike-prow-plugins/pkg/log"
 	"github.com/arquillian/ike-prow-plugins/pkg/scm"
+	"github.com/arquillian/ike-prow-plugins/pkg/utils"
 	"github.com/google/go-github/github"
 )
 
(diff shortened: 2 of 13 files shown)


docs/fragments/work-in-progress.adoc
```diff
@@ -0,0 +1,17 @@
+== Work In Progress plugin
```
bartoszmajsak: Thanks for spending extra effort on documenting it!

pkg/github/status_service.go
```diff
@@ -28,32 +29,33 @@ func NewStatusService(client *github.Client, log log.Logger, change scm.Reposito
 }
 
 // Success marks given change as a success.
-func (s *StatusService) Success(reason string) error {
-	return s.setStatus(StatusSuccess, reason)
+func (s *StatusService) Success(reason, targetURL string) error {
```
bartoszmajsak: I was wondering - since we already have a concept of `StatusContext` which is part of this service, maybe we could have target URLs as part of it? Then we can keep simple signatures with only the `reason` to be passed?

bartoszmajsak:
Also, any idea why your branch builds are failing? https://travis-ci.org/MatousJobanek/ike-prow-plugins/builds/361053342#L712

pkg/github/status_service.go
MatousJobanek: I'm not sure if the `StatusContext` is the best place for it. If we moved it there, then we should move the `reason` there as well. When you take a look at the usage, then you can see that the `TargetURL` is different for each request - every details link that is related only to one reason

MatousJobanek:
> I did few minor wording improvements as per your suggestion and also fixed doc generation.

Thanks a lot for wording improvements :+1: all of them look much better

As for the sentry link - I did it because I think that you have mentioned such a possibility some time ago (but I can be wrong) also there is the comment:
`// TODO create link to detailed log about the problem`
https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/plugin/test-keeper/plugin/test_status_service.go#L47
so, I thought that it was intended to add it there.

I can keep it empty (without the link), I'm fine with it :-)

bartoszmajsak:
> // TODO create link to detailed log about the problem

This `TODO` was the wrong assumption then. Or at least we don't have per-execution log exposed yet.

pkg/github/status_service.go
bartoszmajsak: My main rationale for that is that those links are right now static so we have them at the moment of service instantiation. We have 3 different links we can use within private functions (that's why I said target URL**s** in the original comment).

MatousJobanek: > My main rationale for that is that those links are right now static so we have them at the moment of service instantiation.

you can say the same about the reasons

> We have 3 different links we can use within private functions (that's why I said target URLs in the original comment).

3 different links for the success and one link for failure. I don't understand how would you figure out which "success" link should be used on the level of the `StatusService` or `StatusContext`. You need to provide it along with the reason. 

btw, the `StatusContext` contains `BotName` and `PluginName` - both of them are same for all "set-status" requests. Details link, as well as the reason, is different for every state.

bartoszmajsak: > 3 different links for the success and one link for failure

Right missed that part. Could be early hour ;) Let's stick to what we have now. Maybe giving `targetURL` a different name like `detailsLink`? Also could make it optional, so we can call it with or without it instead of passing `""` in case of latter.

MatousJobanek: renamed here 7043f42 

> Also could make it optional, so we can call it with or without it instead of passing "" in case of latter.

I would keep it mandatory for both statuses failure and success. I don't think that there will be a use case when you need to set the state without providing the details link.

Issue #55, opened 2018-03-21
Bot name should be configurable


bartoszmajsak:
In several places, we are using `ike-plugins` as the name of our tooling. That is also used as `BotName` for Status calls. We should make this configurable the same way as we do with GitHub token (through a flag) and populate it across the plugins.

https://github.com/arquillian/ike-prow-plugins/blob/7543b86822791a367f935fcf22f04b3e35130873/pkg/plugin/test-keeper/plugin/test_status_service.go#L16

PR #83, opened 2018-04-04
fix(#55): makes Bot name configurable.

hemanik:
Makes Bot name configurable.


alien-ike:
### Ike Plugins (test-keeper)

If you are an admin and you are sure that no test is needed then you can use the command `/ok-without-tests` as a comment to make the status green.


For more information please head over to official [documentation](http://arquillian.org/ike-prow-plugins/#_test_keeper_plugin). You can find there how to configure the plugin - for example exclude certain file types so if PR contains only them it won't be checked.

pkg/plugin/test-keeper/plugin/test_status_service.go
```diff
@@ -39,7 +39,7 @@ const (
 )
 
 func (gh *GitHubTestEventsHandler) newTestStatusService(log log.Logger, change scm.RepositoryChange) testStatusService {
-	statusContext := github.StatusContext{BotName: "alien-ike", PluginName: ProwPluginName}
+	statusContext := github.StatusContext{BotName: gh.BotName, PluginName: ProwPluginName}
```
bartoszmajsak: So it was hardcoded to `alien-ike` and now it `plugin_init` we make it 

```
flag.String("bot-name", "ike-plugins", "Bot Name used for the plugin.")
```

Why?

hemanik: Shouldn't the `BotName` be consistent across the plugins? If we want to have hardcoded values then why make it configurable?

bartoszmajsak: Ok I might not be clear in the morning. Sorry, very little sleep this night.

What I meant is we should make it `alien-ike` across all the plugins so the default should be changed :) (have a look at the checks below, `BotName` is first part of the `xxx/yyy` naming we have for the plugins and since we are using @alien-ike account it should be configured to that. For the moment defaulting it is fine.

pkg/plugin/work-in-progress/plugin/event_handler_unit_test.go
```diff
@@ -33,6 +34,7 @@ var _ = Describe("Work-in-progress Plugin features", func() {
 		DescribeTable("should not recognize PR as work-in-progress if title doesn't start with WIP",
 			func(title string) {
 				Expect(handler.IsWorkInProgress(&title)).To(BeFalse())
+				Expect(handler.BotName).To(Equal("alien-ike"))
```
bartoszmajsak: I would think about testing it on `GitHub` level checking if status has proper "context" field instead. WDYT @MatousJobanek?

Here we simply test if constructor works, not sure if that's the optimal route.

pkg/plugin/test-keeper/plugin/test_status_service.go
MatousJobanek: I have discussed it with Hemani yesterday and this was kind of my idea - I understood the issue that all occurrences of the `BotName` should be same and equal to what is set in the flag - sorry for confusions :-)

pkg/plugin/work-in-progress/plugin/event_handler_unit_test.go
MatousJobanek: I do agree with you @bartoszmajsak - this should have been verified on the level of the post request sent by `testStatusService` - @hemanik  take a look here: https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/plugin/test-keeper/plugin/event_handler_gh_test.go#L26

pkg/plugin/test-keeper/plugin/event_handler_gh_test.go
```diff
@@ -55,7 +58,7 @@ var _ = Describe("Test Keeper Plugin features", func() {
 
 			gock.New("https://api.github.com").
 				Post("/repos/" + repositoryName + "/statuses").
-				SetMatcher(ExpectPayload(toBe(github.StatusSuccess, keeper.TestsExistMessage, keeper.TestsExistDetailsLink))).
+				SetMatcher(ExpectPayload(toBe(github.StatusSuccess, keeper.TestsExistMessage, utils.StringJoin(botName, keeper.ProwPluginName), keeper.TestsExistDetailsLink))).
```
bartoszmajsak: how about using a const for expected context name, as this is not changing anywhere in the code?

```golang
const (
  expectedContext = utils.StringJoin(botName, keeper.ProwPluginName)
)
```

pkg/utils/strings.go
```diff
@@ -33,3 +37,8 @@ func convertSliceToInterface(s interface{}) (slice []interface{}) {
 
 	return slice
 }
+
+// StringJoin takes two strings and joins them with defaultConnector
+func StringJoin(prefix string, suffix string) string {
+	return prefix + defaultConnector + suffix
```
bartoszmajsak: Why not https://golang.org/pkg/strings/#Join, e.g. `strings.Join(botName, wip.ProwPluginName, '/')` ?

pkg/utils/strings.go
```diff
@@ -2,6 +2,10 @@ package utils
 
 import "reflect"
 
+const (
+	defaultConnector = "/"
+)
```
MatousJobanek: This is not used anywhere, is it?

Issue #73, opened 2018-03-27
Bypassing test check should be possible for reviewers but not PR creator


bartoszmajsak:
Currently, we only allow admins to bypass test verification by using `/ok-without-tests` command in the PR comments. We should enhance this model to cover following use cases

- reviewers should be able to bypass the check too
- PR author shouldn't be able to do that

https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/plugin/test-keeper/plugin/event_handler.go#L91-L99

- [x] create permission service abstracting this logic away from the event handler (open to decide how to deal with roles, e.g. passing as params etc)
- [x] log as `Warn` an attempt to bypass without proper permission
- [x] comment on the issue in such a case saying: `Sorry @handle but you are not allowed to bypass test-keeper check` (optional and to discuss)
- [x] extend test cases (but that should be assumed and will be caught by the plugin here anyway :P)
- [x] update documentation

bartoszmajsak:
In the reference to https://github.com/arquillian/ike-prow-plugins/pull/85#discussion_r179176020

After a long sleep, I think we should not sacrifice improved user experience (comments indicating if something went wrong) vs availability concerns (e.g. API limits) which at this point of time might be premature. Let's think about different path here:

- prevent from duplicated comment
- log API limit reach
- see how `deck` service from Prow infra can be used to list commands to all the plugins configured per repository (this is what [`PluginHelp`](https://github.com/kubernetes/test-infra/blob/8a2e7d1d10b138997830ec18bf4d8bd142355874/prow/pluginhelp/pluginhelp.go#L40) struct is for)
- when we reach a certain threshold of API limit disable some features  (idea to delve further)
- metrics (this needs to be further evaluated)
  - API calls per PR
  - total GH calls per hour (or limit value per hour)
  - anything else we can think of
- evaluate GH  app approach or user sharding (so we can have one user per org instead of alien-ike for all)

WDYT @MatousJobanek. What else can we improve?

MatousJobanek:
Thank you for great ideas. I love all of them and totally agree that we need some more analysis and mainly the metrics. additional logic for disabling some features when we reach a certain threshold would be awesome.

There is another thing I have in my mind - to take a look at `go-github` library if they use some caching mechanism that would reuse older responses and retrieve headers for making conditional requests: https://developer.github.com/v3/#rate-limiting

PR #85, opened 2018-04-04
chore: refactoring of the comment_service

MatousJobanek:
renamed `comment_service` to `hint_comment_service` and created a new `comment_service` that only adds the given comment. 
The reason is that we will need to have a service (that keeps `RepositoryIssue`) and only adds the given message.  The `hint_comment_service` does a lot of other things...

### DIFF
--- a/pkg/github/comment_service.go
+++ b/pkg/github/comment_service.go
@@ -1,82 +1,16 @@
 package github
 
 import (
-	"fmt"
-	"strings"
-
-	"github.com/arquillian/ike-prow-plugins/pkg/log"
 	"github.com/arquillian/ike-prow-plugins/pkg/scm"
-	"github.com/arquillian/ike-prow-plugins/pkg/utils"
 )
 
-// PluginTitleTemplate is a constant template containing "Ike Plugins (name-of-plugin)" title with markdown formatting
-const (
-	PluginTitleTemplate     = "### Ike Plugins (%s)"
-	assigneeMentionTemplate = "Thank you @%s for this contribution!"
-)
-
-// CommentService is a struct managing plugin comments
+// CommentService is a struct managing issue or pull request comments
 type CommentService struct {
-	client         *Client
-	log            log.Logger
-	issue          scm.RepositoryIssue
-	commentContext CommentContext
+	client *Client
+	issue  scm.RepositoryIssue
 }
 
-// CommentContext holds a plugin name and a assignee to be mentioned in the comment
-type CommentContext struct {
-	PluginName string
-	Assignee   string // TODO rethink this naming when plugins will start interacting with issue creators and reviewers
-}
-
-// NewCommentService creates an instance of GitHub CommentService for the given CommentContext
-func NewCommentService(client *Client, log log.Logger, change scm.RepositoryChange, issueOrPrNumber int, commentContext CommentContext) *CommentService {
-	return &CommentService{
-		client: client,
-		log:    log,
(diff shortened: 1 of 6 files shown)


pkg/github/hint_comment_service.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// PluginTitleTemplate is a constant template containing "Ike Plugins (name-of-plugin)" title with markdown formatting
+const (
+	PluginTitleTemplate     = "### Ike Plugins (%s)"
+	assigneeMentionTemplate = "Thank you @%s for this contribution!"
+)
+
+// HintCommentService is a struct managing plugin comments
+type HintCommentService struct {
```
bartoszmajsak: I think simple `HintService` (or even `Hinter`) would do too.

pkg/github/hint_comment_service.go
```diff
@@ -0,0 +1,81 @@
...
+	*CommentService
+	log            log.Logger
+	commentContext HintCommentContext
+}
+
+// HintCommentContext holds a plugin name and a assignee to be mentioned in the comment
+type HintCommentContext struct {
```
bartoszmajsak: Same here then `HintContext`

pkg/github/comment_service.go
```diff
@@ -1,82 +1,16 @@
...
-	return s.append(s.getPluginTitle(), fmt.Sprintf(assigneeMentionTemplate, s.commentContext.Assignee))
+	client *Client
+	issue  scm.RepositoryIssue
 }
 
-func (s *CommentService) getPluginTitle() string {
-	return fmt.Sprintf(PluginTitleTemplate, s.commentContext.PluginName)
+// Comment adds a comment message to the issue
+func (s *CommentService) Comment(commentMsg *string) error {
+	return s.client.CreateIssueComment(s.issue, commentMsg)
```
bartoszmajsak: If the only thing this service does now is to delegate to `CreateIssueComment` from our `go-github` facade, do we really need it?

bartoszmajsak: > plugin should comment that a user is not allowed to use some command

Not sure if we shouldn't just ignore such case. Will easily become a problem of running out of API call limits.

pkg/github/hint_comment_service.go
MatousJobanek: `Hinter` is cool :+1:

pkg/github/comment_service.go
MatousJobanek: well, I think that the number of the not-allowed attempts won't be too much. And, from my point of view, it is much better to notify/comment why the command is not working for such a user than just to stay silent. The current comment msg that is sent looks like:

> @bartoszmajsak-test has used a command `/ok-without-tests` but this user is not allowed to do that. User with the necessary permissions is anybody who is admin or requested reviewer, but not pull request creator

This comment is generated automatically based on the information that are set for the specific command...

bartoszmajsak: We have 5k API calls per hour, right? Multiply it per repos, prs and edits on them. I would be very careful with extra comments, due to this fact. It also can be gamed by anyone having account on GH and we can easily put the service down in no time.

If we list all possible commands and roles allowed in the hint in `<details>` section, for example, we have it all explained without a need of repeating it over and over again.

bartoszmajsak: Moved the discussion here https://github.com/arquillian/ike-prow-plugins/issues/73#issuecomment-378845901

PR #86, opened 2018-04-04
fix(#87): retry function exits when response's exit code is < 404 and returns error that is not nil

MatousJobanek:
* there was a bug that when the response returned an error and the exit code was lower than 404, then it didn't return the error.
* another bug was in exit code formatting:
`return errors.New("Server responded with error " + string(response.StatusCode))`
as a result, the output contained:
`1. [Server responded with error Ɣ]`
* I added also test cases for it

Fixes: #87

### DIFF
--- a/pkg/http/retry.go
+++ b/pkg/http/retry.go
@@ -11,21 +11,24 @@ import (
 // RequestInvoker is a function that invokes a request and returns a response
 type RequestInvoker func() (*gogh.Response, error)
 
-// Do invokes a request for the given amount of retries with the given sleep before each one of them until the response status code is lower than 404
+// Do invokes a request for the given amount of retries with the given sleep before each one of them until the response
+// status code is lower than 404 or the returned error is not nil
 func Do(retries int, sleep time.Duration, invokeRequest RequestInvoker) error {
-	errs := utils.Retry(retries, sleep, func() error {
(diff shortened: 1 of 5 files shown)


bartoszmajsak:
@MatousJobanek can you create a bug and link to it in the PR title?

pkg/http/retry.go
```diff
@@ -11,21 +11,24 @@ import (
...
 			}
-			return errors.New("Server responded with error " + string(response.StatusCode))
+			return true, fmt.Errorf("server responded with error %d", response.StatusCode)
 		}
-		return nil
+		return false, err
 	})
 	if len(errs) > 0 {
-		msg := fmt.Sprintf("During %d retries of sending a request %d errors have occurred:", retries, len(errs))
+		msg := fmt.Sprintf("During %d retries of sending a request %d errors have occurred:", performedRetries, len(errs))
```
bartoszmajsak: Minor thing, but maybe we can make an error msg like this:

`%d out of %d attempts failed` 

plus maybe what we actually tried to call if it's easily accessible?

MatousJobanek: actually, if the log is logged then it means that all attempts have failed. so maybe the better message would be:
`All %d attempts of sending a request failed. See the errors:`

PR #88, opened 2018-04-06
chore(#36): brings cmd/ folder for all main.go plugins

### LINKED ISSUE #36: Flatten plugin/ directories

bartoszmajsak:
Based on @MatousJobanek suggestion https://github.com/arquillian/ike-prow-plugins/pull/64#issuecomment-376169125 with the difference that we keep `cmd/main.go` as this is a command, and not shareable package.

Let's keep it open until we merge other PRs as it moves quite a few files around. I will take care of merging `master` on the way here.

Fixes #36


MatousJobanek:
Don't forget to also:
- [ ] change paths in docs

bartoszmajsak:
Sure, I have it on my list, but thanks for reminding.

PR #89, opened 2018-04-06
feat(#73): bypassing test check is possible also for reviewers but not PR creator

MatousJobanek:
* restricted bypassing test check to admin or requested reviewer but not PR creator
* added new abstractions:
  * `CommentCmd` is a abstraction of particular comment command. Has two methods:
    * `Perform` that evaluates and executes the command logic itself
    * `Matches` that returns true when comment content matches the expected command
  * `CommentCmdHandler` that can register multiple `CommentCmd` implementations and then handle a comment event against these commands
    * performs only those comment commands that match the comment content
  * `CmdExecutor` - provides an API for setting restrictions and providing a function that should be executed when the restrictions are fulfilled. When the permission restrictions are not met, then it logs a `WARN` with the generated message and if the mode is not quite, then also adds the comment to the PR. Restrictions that can be set are:
     * actions - for which events the command should be triggered - eg.: `When(is.Triggered)`
     * permissions - who can trigger the command - eg.:`By(is.AnyOf(user.ThatIsAdmin, user.ThatIsPRReviewer), is.Not(user.ThatIsPRCreator))`
* Added lazy loading of PRs - see `PullRequestLoader`. The reason is to avoid sending requests when they are not needed (comment content or the actions are not matched). So the PR information is loaded only when needed and only once.
* added tests and changed docs
* changed time sleep in default GH client used in tests to one nanosecond - this will make the tests a bit faster :-)

Note: I know that the naming is probably are not the best - I'm open to any proposals. Complications that I was facing were:
* I had to name the directory `command` because `cmd` is for `main.go` file
* When I have `command` directory then the structs cannot start with `Command` as I had originally, so that's why they contains `Cmd` instead
* reason for naming the import
 `is "github.com/arquillian/ike-prow-plugins/pkg/command"`
 is fluent reading of the code

Fixes: #73


pkg/command/permission_service_test.go
```diff
@@ -0,0 +1,223 @@
...
+				PRLoader: &github.PullRequestLazyLoader{
+					Client:    client,
+					RepoOwner: "owner",
+					RepoName:  "repo",
+					Number:    1,
+				},
+			}
+		})
+
+		It("tests 'that is admin' function that should not approve the user when the permission is read", func() {
```
bartoszmajsak: Do we really need `tests 'that is admin' function that ` prefixes in those tests?

pkg/command/permission_service_test.go
```diff
@@ -0,0 +1,223 @@
...
+				Get("/repos/owner/repo/pulls/1").
+				Reply(200).
+				BodyString(`{"user": {"login": "user"}}`)
+
+			// when
+			permissionStatus, err := user.PRCreator()
+
+			// then
+			Ω(err).ShouldNot(HaveOccurred())
+			verifyStatusWithPredefinedUser(permissionStatus, true, "pull request creator")
```
bartoszmajsak: Would it be possible (if quick enough) to make it a custom assertion `Expect(....)` (probably with `SatisfyAll`)?

pkg/command/permission_service.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func (s *PermissionStatus) allow() *PermissionStatus {
+	s.UserIsApproved = true
+	return s
+}
+
+func (s *PermissionStatus) constructMessage(operation, command string) string {
+	msg := fmt.Sprintf(
+		"@%s has %s a command `%s` but this user is not allowed to do that. "+
```
bartoszmajsak: > but this user is not allowed to do that. 

how about

> but this will not have any effect due to insufficient permission

Plus maybe log users' permission too?

pkg/github/pr_lazy_loader_test.go
```diff
@@ -0,0 +1,59 @@
...
+	It("should load pull request when load() method is called", func() {
+		// given
+		counter := 0
+		gock.New("https://api.github.com").
+			Get("/repos/owner/repo/pulls/123").
+			SetMatcher(createCounterMather(&counter)).
+			Reply(200).
+			BodyString(`{"title":"Loaded PR"}`)
+		loader := &github.PullRequestLazyLoader{Client: client, RepoOwner: "owner", RepoName: "repo", Number: 123}
+		Expect(counter).To(Equal(0))
```
bartoszmajsak: Do we need that assertion here?

pkg/internal/test/test_doubles.go
```diff
@@ -47,7 +49,13 @@ func ExpectPayload(payloadAssert func(payload map[string]interface{}) bool) gock
 		}
 		var payload map[string]interface{}
 		err = json.Unmarshal(body, &payload)
-		payloadExpectations := payloadAssert(payload)
+		var payloadExpectations bool
+		for _, assertion := range payloadAssert {
+			payloadExpectations = assertion(payload)
+			if !payloadExpectations {
+				break
```
bartoszmajsak: I was wondering - would it make sense to make it a "soft assertion" instead?

pkg/internal/test/test_doubles.go
```diff
@@ -66,6 +74,13 @@ func NewDefaultGitHubClient() *github.Client {
 	return &github.Client{
 		Client:  gogh.NewClient(nil), // TODO with hoverfly/go-vcr we might want to use tokens instead to capture real traffic
 		Retries: 3,
-		Sleep:   time.Second,
+		Sleep:   time.Nanosecond,
```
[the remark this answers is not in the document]
MatousJobanek: as it is done in your PR, sure I will

pkg/command/permission_service_test.go
MatousJobanek: it seems that it doesn't have to be there :-D. To be honest, I have no idea why I added it there. The reason was that when I combined the text with the name of the function `It`, then it didn't sound natural to me:
`it - should not approve the user when the permission is read"`
but I have no idea why :-D. 
I'll remove it ;-)

pkg/command/permission_service.go
MatousJobanek: sure, that sounds better

pkg/github/pr_lazy_loader_test.go
MatousJobanek: I just wanted to verify that it wasn't invoked before the actual call of `loader.Load`.

pkg/internal/test/test_doubles.go
MatousJobanek: the same came to my mind I wanted to do it but forgot :-D

pkg/command/permission_service_test.go
MatousJobanek: I'll try my best

pkg/internal/test/test_doubles.go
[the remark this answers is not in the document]
bartoszmajsak: Right :D I wasn't sure which one comes first :)

pkg/internal/test/test_doubles.go
bartoszmajsak: Sure, I need this DSL :)

pkg/plugin/test-keeper/plugin/comment_cmd.go
```diff
@@ -0,0 +1,41 @@
...
+	"strings"
+
+	is "github.com/arquillian/ike-prow-plugins/pkg/command"
+	"github.com/arquillian/ike-prow-plugins/pkg/github"
+	gogh "github.com/google/go-github/github"
+	"github.com/arquillian/ike-prow-plugins/pkg/log"
+)
+
+// SkipComment is used as a command to bypass test presence validation
+const SkipComment = "/ok-without-tests"
```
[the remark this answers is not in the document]
MatousJobanek: renamed a84f6f4

pkg/plugin/test-keeper/plugin/comment_cmd.go
```diff
@@ -0,0 +1,41 @@
...
+
+// SkipCommentCmd represents a command that is triggered by "/ok-without-tests"
+type SkipCommentCmd struct {
```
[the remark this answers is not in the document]
MatousJobanek: renamed a84f6f4

pkg/internal/test/test_doubles.go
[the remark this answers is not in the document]
MatousJobanek: done in fa0f8bd

pkg/command/permission_service_test.go
MatousJobanek: removed in 0c91c60

pkg/command/permission_service_test.go
MatousJobanek: will be part of https://github.com/arquillian/ike-prow-plugins/issues/105

pkg/command/permission_service.go
MatousJobanek: implemented/changed in 20d461e

pkg/internal/test/test_doubles.go
MatousJobanek: will be part of #105

PR #94, opened 2018-04-09
fix(#92): excludes internal config files from test verification

### LINKED ISSUE #92: Test Keeper configuration not respected when included in the PR
https://github.com/arquillian/ike-prow-plugins/pull/82 brings new repository into `osio-cluster.yml`. Such PRs should not be verified for test presence so I tried to add `test-keeper.yml` make this file excluded from the verification process, but it apparently does not work as expected.

bartoszmajsak:
- fixes a bug where the configuration is added/changed as part of the PR, but it's treated as a regular file instead of being excluded from the verification
- brings minor improvements for `**/...` handling 

Fixes #92


pkg/assets/config/test-keeper.yaml
```diff
@@ -99,4 +99,8 @@ skip_validation_for:
 
   # Shell
   - '*.sh'
-  - '*.bat'
\ No newline at end of file
+  - '*.bat'
+
+  # Our own configurations
+  - 'test-keeper.yaml'
+  - 'test-keeper.yml'
```
bartoszmajsak: @MatousJobanek this probably is a very strong argument to move all our internal configuration, hints et al to single location like `.ike-prow/` - would make it one line ;) Issue coming.

MatousJobanek: I can only agree with you, so +1 from my side

pkg/plugin/test-keeper/plugin/file_pattern.go
```diff
@@ -51,18 +66,20 @@ func parseFilePattern(pattern string) string {
...
+		expr += endOfLineRegexp
 	}
 
-	return regexp
+	return expr
 }
 
 func transformPathPatternToRegexp(path string) string {
+	for ; strings.HasPrefix(path, anyPathWildcard+"/"); path = path[strings.Index(path, "/")+1:] {
+	}
```
MatousJobanek: I would change it to
```java
	for strings.HasPrefix(path, anyPathWildcard+"/") {
		path = path[strings.Index(path, "/")+1:]
	}
```

Issue #90, opened 2018-04-06
Remove Fatal method from the logger and replace usages by Error


MatousJobanek:
When `Fatal` log level is used, then logrus exits the process which makes very hard figuring out which test is failing. In addition, there is no need to exit the process except for the cases when secrets are not loaded in `plugin_init`

PR #96, opened 2018-04-10
fix(#90): relaxes logging to error for PR handling logic

bartoszmajsak:
Otherwise, we exit the process. Fatal should only be used when we really cannot continue running any plugin.

### DIFF
--- a/pkg/plugin/test-keeper/plugin/event_handler.go
+++ b/pkg/plugin/test-keeper/plugin/event_handler.go
@@ -90,7 +90,7 @@ func (gh *GitHubTestEventsHandler) handlePrComment(log log.Logger, prComment *go
 	sender := prComment.Sender.Login
 	permissionLevel, e := gh.Client.GetPermissionLevel(*org, *name, *sender)
 	if e != nil {
-		log.Fatal(e)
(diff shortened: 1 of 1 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @bartoszmajsak for this contribution!



If you are an admin and you are sure that no test is needed then you can use the command `/ok-without-tests` as a comment to make the status green.


Your plugin configuration is stored in the [file](https://github.com/arquillian/ike-prow-plugins/db822903a84a0616f993e385733b0819b78b0429/test-keeper.yaml).

pkg/log/log.go
```diff
@@ -7,15 +7,13 @@ import (
 // Logger is a facade interface compatible with logrus.Logger but with limited scope of logging.
 // It exists to decouple plugin implementations from particular log implementation but also to only allow
 // reporting actionable problems and corner cases such as panic (or in other words to avoid logging as program flow analysis / debugging).
+// If needed this can be extended by adding other levels such as Panic or Fatal (both are exiting, former go-routine or the program if unwinding reaches
```
MatousJobanek: I like this note!

Issue #105, opened 2018-04-11
Add soft matcher and additional custom assertion


MatousJobanek:
- [ ] implement soft matcher that will behave as a soft assertion in AssertJ
- [ ] add a custom assertion for `PermissionStatus` object

PR #106, opened 2018-04-11
feat(#105): adds custom matcher for PermissionStatus and soft matcher for soft assertions

MatousJobanek:
* soft matcher that can be used eg. via function `SoftlySatisfyAll(matchers ...SoftMatcher)`
  * `SoftMatcher` struct creates a message that is then incorporated into the failure message
  * `NamedGomegaMatcher` is a matcher that also keeps a name of the element that is checked (this is included in the failure message)
  * `SoftlyAllMatcher` keeps an array of matchers to be matched and divides them into two arrays based on the information if the match was successful or not
* custom matcher for `PermissionStatus` that is using the soft assertions. The usage example is:
```go
ExpectPermissionStatus(status).To(
	HaveApprovedUser("user"),
	HaveApprovedRoles(is.RequestReviewer),
	HaveNoRejectedRoles())
```
failure message example is:
```
    For the given input:
    
        <command.PermissionStatus>: {
            User: "user",
            UserIsApproved: false,
            ApprovedRoles: ["admin"],
            RejectedRoles: nil,
        }
    
    the following matchers failed:
    
    1) [User]
    Expected
        <string>: user
    to equal
        <string>: developer
    
    2) [UserIsApproved]
    Expected
        <bool>: false
    to equal
        <bool>: true
    
    3) [ApprovedRoles]
    Expected
        <[]string | len:1, cap:1>: ["admin"]
    to be empty
```

* moved all matchers related to payload to the file `payload_matchers.go`
  * changed the logic so it is using the soft assertions and also to simplify the usage so currently it is not necessary to pass to the `ExpectPayload` some complicated function but just a list of `SoftMatchers`.

failure message example is:
```
    For the given input:
    
        <map[string]interface {} | len:4>: {
            "state": "success",
            "target_url": "http://arquillian.org/ike-prow-plugins/#tests-exist",
            "description": "There are some tests :)",
            "context": "alien-ike/test-keeper",
        }
    
    the following matchers failed:
    
    1) [state]
    Expected
        <string>: success
    to equal
        <string>: failure
    
    2) [description]
    Expected
        <string>: There are some tests :)
    to equal
        <string>: Failed while check for tests
```

Fixes: #105


pkg/internal/test/soft_matcher.go
```diff
@@ -0,0 +1,128 @@
...
+			m.failedMatchers = append(m.failedMatchers, matcher)
+		} else {
+			m.passedMatchers = append(m.passedMatchers, matcher)
+		}
+		if err != nil {
+			return false, err
+		}
+	}
+
+	return len(m.failedMatchers) == 0, nil
```
bartoszmajsak: Will this ever be different than `0` when reaching this line of code? if so shouldn't we have an `err` here too? Or is `err` not necessary carrying "assertion error" information?

pkg/internal/test/soft_matcher.go
```diff
@@ -0,0 +1,128 @@
...
+
+	for _, matcher := range m.Matchers {
+		success, err := matcher.Match(actual)
+		if !success {
+			m.failedMatchers = append(m.failedMatchers, matcher)
+		} else {
+			m.passedMatchers = append(m.passedMatchers, matcher)
+		}
+		if err != nil {
+			return false, err
```
bartoszmajsak: Shouldn't we continue here and collect all `err`?

bartoszmajsak: Or is it considered a "terminal state" regardless?

MatousJobanek:
> Do you think we could contribute this upstream at some point?

I was thinking about it from the very beginning - that this should be part of the upstream. Let's have it for some time only in our test suite to check if there is something missing or not. When we know that it will not require some bigger changes in the code, then send it as a PR to upstream

pkg/internal/test/soft_matcher.go
bartoszmajsak: Ok, that's what I wanted to understand, fair enough.

PR #107, opened 2018-04-11
feat(#91): gock tests improvements

### LINKED ISSUE #91: Check if all gock requests have been matched
In functional tests, we use gock library for mocking GH requests. Sometimes it can happen that a request is not properly matched and in a combination with wrong error handling, we don't get any failure. 
We should check if all requests have been matched and all mocked requests have been used in the test:
```go
AfterEach(func() {
	Expect(gock.GetUnmatchedRequests()).To(BeEmpty())
	Expect(gock.GetAll()).To(BeEmpty())
})
```

bartoszmajsak:
This PR adds `AfterEach` call to make sure no unmatched requests were made during the gock tests.

```golang
gomega.Expect(gock.GetUnmatchedRequests()).To(gomega.BeEmpty(), "Have no unmatched requests")
```

In addition `BeforeEach` calls `defer gock.OffAll()` because we want to purge the registry of unmatched requests to start from the clean state for every test.

Original proposal was to have following two checks:

```go
AfterEach(func() {
	Expect(gock.GetUnmatchedRequests()).To(BeEmpty())
	Expect(gock.GetAll()).To(BeEmpty())
})
```

however

> GetAll returns the current stack of registed mocks.

I didn't add this option, because after every test we still have all the mocks kept - they are removed before every other test.

Fixes #91


pkg/internal/test/gock.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/onsi/gomega"
+	"gopkg.in/h2non/gock.v1"
+)
+
+// EnsureGockRequestsHaveBeenMatched checks if all requests have been matched and all mocked requests have been used in the test
```
MatousJobanek: it just checks if all requests have been matched, but not that all mocked requests have been used in the test - that was the function of `GetAll`

bartoszmajsak: Correct, the docs were for the original version with both checks

bartoszmajsak:
If I add `GetAll` check it fails with some gargantuan mock response logged
which seems to be called in the test anyway

This can mean two things:
- the requests are not removed when they are matched (persisted by default)?
- we call the same endpoint more than once

Seems to be the first case https://gist.github.com/bartoszmajsak/0bfa283ffb450bd87e8ddac3f29a151d#L101

Issue #67, opened 2018-03-27
Add common exclusion patterns


bartoszmajsak:
Add more common files extensions/patterns which will be used to determine if PR should be verified against test presence. If only those files are present in the changeset it should be skipped.

- common image formats
- configuration for GitHub apps used among OSIO services (e.g. Codecov)
- dependency tools we don't cover yet (e.g. godep)

PR #109, opened 2018-04-12
fix(#67): extends default exclusion patterns to verify against test presence.

hemanik:
Extends exclusion patterns list with few common exclusion files to be skipped for test presence.


bartoszmajsak:
As discussed offline, there are few things still to be added before the review, thus adding WIP to the title.

@hemanik please remove when you add those.

PR #111, opened 2018-04-12
feat(#110): changes deployments to deploymentconfigs

### LINKED ISSUE #110: Change Deployment objects to DeploymentConfigs
We are currently using standard k8s `Deployment` objects. Using Openshift `DeploymentConfigs` we can add triggers and customize `Rolling` update strategy.

Other OSIO services are using this approach as well.

bartoszmajsak:
- switches to OpenShift DeploymentConfigs as other OSIO services
- introduces handy make targets to restart deployments (e.g. when
configmap or secret changes)
  - `oc-restart-all` restarts hook service and all deployment plugins
  - `oc-restart-hook` restarts just a hook service
  - `oc-restart-plugins` restarts all deployed plugins

**NOTE**: If you have services already deployed as Deployment objects, run following script to remove them. Otherwise, you will end up having two (or more) pods of the same service

```shell
oc get deployments --no-headers=true -o=custom-columns=NAME:metadata.name | xargs -I {} oc delete deployments/{}
```

Fixes #110


alien-ike:
### Ike Plugins (test-keeper)

Thank you @bartoszmajsak for this contribution!



Your plugin configuration is stored in the [file](https://github.com/arquillian/ike-prow-plugins/56f6c2069b689c3962046a50393020175b518263/test-keeper.yaml).

PR #112, opened 2018-04-13
feat: adds more exclusion patterns

xcoulon:
Ignoring `dep` files, `.dockerignore` and supporting different names for the `docker-compose` files (eg: `docker-compose.dev.yml` or `docker-compose.macos.yml` if there are specificities per host/env in a project.


bartoszmajsak:
@xcoulon we haven't really reached the goal of the PR description :) Turned out that pattern for `docker-compose` was wrong. If only we would have tests here and plugin would actually catch that ;)

See https://github.com/arquillian/ike-prow-plugins/pull/125

xcoulon:
ah, thanks for catching that bug, @bartoszmajsak ! (and sorry for the late response, too!)

Issue #98, opened 2018-04-10
Create centos-based images for our plugins


bartoszmajsak:
In order to host our services in OSIO clusters we have to  based our container images on Centos (later RHEL). See sample docker files from other OSIO services.

For our own plugins it's fairly easy task, we have to check what is coming from the git-prow images we are inheriting from.

In case of `hook` services which is our dependency from `k8s/test-infra` project we have to figure out the easiest way of building the image without being behind the upstream images (and code base).

bartoszmajsak:
For RHEL based images these are places to look at

- https://github.com/fabric8-services/fabric8-auth/pull/432
- https://github.com/openshiftio/openshiftio-cico-jobs/pull/589

PR #114, opened 2018-04-16
feat(#98): migrates images to use centos

MatousJobanek:
* the docker images are using `centos:7` instead of  `gcr.io/k8s-prow/git:0.1`
* hook binary is built locally
* version of the hook image reflects the prow version set in `glide.yaml`

Fixes: #98


Makefile
```diff
@@ -15,6 +15,7 @@ PUSH_IMAGES:=$(patsubst %,push-%, $(PLUGINS))
 CLEAN_IMAGES:=$(patsubst %,clean-%, $(PLUGINS))
 OC_DEPLOYMENTS:=$(patsubst %,oc-%, $(PLUGINS))
 OC_RESTART:=$(patsubst %,dc-%, $(PLUGINS))
+PROW_VERSION=$(shell grep "k8s.io/test-infra/prow/hook" ${PWD}/glide.yaml -A 1 | grep version | awk '{print $$2}')
```
bartoszmajsak: `:=` should do :)

Makefile
```diff
@@ -125,16 +130,33 @@ oc-deploy-starter: ## Deploys basic prow infrastructure
...
 
-HOOK_VERSION?=v20180403-2f3ad698f
 .PHONY: oc-deploy-hook
-oc-deploy-hook: ## Deploys hook service only
-	@echo "Deploys hook service ${HOOK_VERSION}"
+oc-deploy-hook: build-hook build-hook-image push-hook-image ## Deploys hook service only
+	@echo "Deploys hook service ${PROW_VERSION}"
 	@oc process -f $(CLUSTER_DIR)/hook-template.yaml \
-    		-p VERSION=$(HOOK_VERSION) \
+			-p REGISTRY=$(REGISTRY) \
```
bartoszmajsak: Minor detail: formatting.

Makefile
```diff
@@ -125,16 +130,33 @@ oc-deploy-starter: ## Deploys basic prow infrastructure
 	@echo "Deploying prow infrastructure"
 	@oc apply -f cluster/starter.yaml
```
bartoszmajsak: Maybe we kick-out the "starter.yaml" entirely?

cluster/hook-template.yaml
```diff
@@ -3,12 +3,12 @@ apiVersion: v1
...
   required: true
-  value: gcr.io
+  value: docker.io
 - name: DOCKER_REPO
   required: true
-  value: k8s-prow
+  value: arquillian
 - name: VERSION
-  value: v20180403-2f3ad698f
+  value: latest
```
bartoszmajsak: We don't publish "latest" so that default won't work. But we always pass it to the template anyway `-p VERSION=$(PROW_VERSION)` so I think we can just remove the default value

Makefile
MatousJobanek: good catch, thx

Makefile
MatousJobanek: ok, I'll remove it. You are right that we don't use it at all so there is no point in keeping it there. In addition, these images are not centos based...

cluster/hook-template.yaml
MatousJobanek: the same applies also to `ike-prow-template.yaml` right? Since this is the place I took it from :-)

bartoszmajsak: Yeah, that's correct. We want to use exact build tag.

Makefile
MatousJobanek: 13f64a2

Makefile
MatousJobanek: b91a1ba

Makefile
MatousJobanek: c629ccf

cluster/hook-template.yaml
MatousJobanek: 673ecf5

PR #121, opened 2018-04-18
chore(#95): expose hook version to openshift

### LINKED ISSUE #95: Expose hook version to OpenShift
Similarly to how we expose versions of plugin images, create it for hook too.

  * label for the `Deployment` https://github.com/arquillian/ike-prow-plugins/blob/bab7aa31f1d6710b976a39783d34bf39d4f5484c/cluster/ike-prow-template.yaml#L18-L23
  * environment variable for the container itself
https://github.com/arquillian/ike-prow-plugins/blob/bab7aa31f1d6710b976a39783d34bf39d4f5484c/cluster/ike-prow-template.yaml#L44-L46

Have a look into `Makefile` how those are passed.

dipak-pawar:
This exposes `VERSION` as env variable in a container and adds version lable for DeploymentConfig.

Fixes #95


bartoszmajsak:
@dipak-pawar could you add description to the PR and reference an issue?

dipak-pawar:
@bartoszmajsak Are we using this exposed env variable in container currently or is this future plan?

bartoszmajsak:
@dipak-pawar It's to know which version is running currently - we have it for each pod in OSO already. Then we can leverage it further if needed, but it's more now to know what is built and deployed - especially with the new approach with centos based images #114

bartoszmajsak:
Thanks @dipak-pawar

Issue #97, opened 2018-04-10
Configure CentosCI builds


bartoszmajsak:
Similarly to other OSIO projects we should build our images on CentosCI and publish it to `devshift` (we can still keep travis build for the time being due to doc generation).

For reference see [Launcher Backend](https://github.com/fabric8-launcher/launcher-backend/) cico scripts.

Ultimately we have to send PR here https://github.com/openshiftio/openshiftio-cico-jobs

MatousJobanek:
I'm reopening this issue as we need also a script/job for pushing images to devshift.

- [ ] create `cico_build_deploy.sh` that builds the project and deploys docker images to `push.registry.devshift.net`
- [ ] tag every version to enable usage of the "latest" tag (will be default in the templates)

PR #123, opened 2018-04-19
chore(#97): introduces CentosCi builds

MatousJobanek:
adds:
* `cico_run_tests.sh` script that prepares the environment execute the whole build within a docker image
* `cico_setup.sh` script containing all functions - common thing for all cico scripts
* `Makefile.docker` that is included in main `Makefile`. Contains methods for building and running docker image containing `ike-prow-plugin` project. Provides also an option to execute any makefile target inside of the running container by just using the prefix `docker-` before the target name itself
* `Dockerfile` that represents environment the project could be built in.
* slightly modified `setup.sh` to enable installing only the go,gvm and glide binaries

to try it locally run: 
```bash
make docker-start
make docker-tools
make docker-install
make docker-build
```

Fixes: #97


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/arquillian/ike-prow-plugins/19d3f963273bace033dbb2abdc0dc1565781d1bd/test-keeper.yaml).

Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM centos:7
```
bartoszmajsak: I think I messed up naming originally.

This file should be named `Dockerfile.builder` as it builds the binaries in docker container, and the current `Dockerfile.builder` we can name `Dockerfile.prow`. WDYT?

This will impact original `Makefile` too

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+LABEL maintainer="Devtools <devtools@redhat.com>"
+LABEL maintainer="Devtools-test <devtools-test@redhat.com>"
+
+RUN yum install -y git mercurial make binutils bison gcc glibc-devel which
```
bartoszmajsak: I doubt we need `mercurial` here. WIT removed it lately.

Makefile
```diff
@@ -23,6 +23,10 @@ DOCKER?=$(if $(or $(in_docker_group),$(is_root)),docker,sudo docker)
 
 .DEFAULT_GOAL := all
 
+CUR_DIR = $(shell pwd)
+
+include ./Makefile.docker
```
bartoszmajsak: Do you think splitting current `Makefile` would help?

I can think of slicing it in the following way

- build
- deployment
- docker-plugins
- docker-builder

If we do it how about moving all those files to `.make` directory?

setup.sh
```diff
@@ -3,38 +3,57 @@
...
+function install_packages() {
+  echo -e "${CLEAR}${LIGHT_GREEN}Install prerequisites${CLEAR}"
 
-# Ubuntu
-if [ -n "$(command -v apt-get)" ]; then
-    sudo apt-get -y install curl git mercurial make binutils bison gcc build-essential
-fi
+  # Fedora
+  if [ -n "$(command -v dnf)" ]; then
+    sudo dnf -y install curl git mercurial make binutils bison gcc glibc-devel which
```
bartoszmajsak: Do we really need all of those packages?

setup.sh
```diff
@@ -3,38 +3,57 @@
...
 
-echo -e "${CLEAR}${LIGHT_GREEN}Installing Go 1.4 first - see ${CLEAR}https://github.com/moovweb/gvm/issues/124${CLEAR}"
-gvm install go1.4 --binary
-gvm use go1.4
+  if [[ -n ${GOPATH} ]]; then
+    ORIGINAL_GOPATH=${GOPATH}
+  fi
+  echo -e "${CLEAR}${LIGHT_GREEN}Installing GVM${CLEAR}"
+  curl -s -S -L https://raw.githubusercontent.com/moovweb/gvm/master/binscripts/gvm-installer | sh
+  [[ -s "${GVM_ROOT}/scripts/gvm" ]] && source ${GVM_ROOT}/scripts/gvm || source ~/.gvm/scripts/gvm # otherwise gvm use will not be recognized
```
bartoszmajsak: I introduced `gvm` for local dev setup - to make it easy to install and switch between the versions. I am not sure if we need that for "builder"  image though. Maybe we can simplify it like they did in [wit](https://github.com/fabric8-services/fabric8-wit/blob/02a2585786e7f74b4bafd1a175a835cdcde6b4d5/Dockerfile.builder)?

bartoszmajsak: Maybe we split it to `setup.sh` and `setup.local.sh`?

Makefile
bartoszmajsak: If so we should add it to our `test-keeper.yml` config.

Dockerfile
MatousJobanek: make sense. I have changed the names several times - I had also:
`Dockerfile.builder` for this one
`Dockerfile.binary` for the prow binaries, but your name is fine too

Dockerfile
MatousJobanek: ok, I just copy-pasted it from the setup

Makefile
MatousJobanek: sounds good. how about these names:
`Makefile.build`
`Makefile.deploy`
`Makefile.docker.binary` or `Makefile.docker.prow` or just `Makefile.docker.plugins`
`Makefile.docker.build`

setup.sh
MatousJobanek: I don't know - It has been already there :-p 
I'll try to minimize the stack ;-)

Makefile
bartoszmajsak: Opting for `Makefile.docker.prow`

Dockerfile
bartoszmajsak: Yeah, I think we can make sure we have consistent naming with other projects, so we won't have confusion. Thanks!

setup.sh
bartoszmajsak: I know it would be ideal to have it all the same, but OTOH gvm part is a developers convenience tool, not really something we want to have in docker images b/c we will never use it.

I think let's have one setup for both, but dropping gvm (maybe document it or make a sh script `install_go_gvm.sh` for convenience). WDYT?

Dockerfile
MatousJobanek: renamed 87811af

Dockerfile
MatousJobanek: minimized the stack of packages as part of 3817dbe

setup.sh
MatousJobanek: minimized the stack of packages as part of 3817dbe

setup.sh
MatousJobanek: removed gvm as part of 3817dbe
I would prefer to have one consistent way of doing it. In addition, the GVM tool didn't seem to be very convenient as it resets the GOPATH variable and requires the installation of another go version. The setup.sh script now contains the downloading logic that can be used both locally and in the Dockerfile.

bartoszmajsak:
@MatousJobanek great work, thanks a lot. 

I took liberty to do some minor changes, so please have a look at my latest commits. There was also one small issue with wrong Dockerfile being used when building hook image (see https://github.com/arquillian/ike-prow-plugins/pull/123/commits/b9137dc2ace624c1d50ae2a655c08e5eb978abd2). Other than that all good to go, so you can merge if you are good with my changes. Feel free to revert whatever does not make sense.

MatousJobanek:
Hi @bartoszmajsak 
thanks a lot for the changes and mainly for fixing the bug - I forgot to change also that line after the huge rename/refactoring change. 
All your changes look good, so I'm going to merge it. Thanks a lot for your review and for your help.

Issue #62, opened 2018-03-23
Add probes to plugins


bartoszmajsak:
Each plugin should provide [readiness and liveness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/) (both can point to the same resource)

For this purpose extend the server to respond to HTTP GET `/version` returning plugin name and its version.

MatousJobanek:
closed via https://github.com/arquillian/ike-prow-plugins/pull/128

PR #128, opened 2018-04-20
feat(#62): adds health check probes to plugins.

hemanik:
Each plugin provides a liveliness and readiness probe by extending the server to respond to HTTP `GET /version` endpoint.


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/arquillian/ike-prow-plugins/3ba097857820bdbf86686ea1364ed8c20c8ab034/test-keeper.yaml).

pkg/plugin/probes_handler.go
```diff
@@ -0,0 +1,17 @@
...
+	"log"
+	"net/http"
+)
+
+// NewProbesHandler registers liveliness and readinesss probes for /version endpoint
+func NewProbesHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		_, err := w.Write([]byte("200"))
+		if err != nil {
```
bartoszmajsak: Can we return plugin version as part of the response and have simple http test for that too?

cluster/ike-prow-template.yaml
```diff
@@ -71,6 +71,18 @@ objects:
...
+                port: 8888
+              initialDelaySeconds: 5
+              periodSeconds: 5
+            livenessProbe:
+              httpGet:
+                path: /version
+                port: 8888
+              initialDelaySeconds: 5
+              periodSeconds: 5  
           volumes:
```
bartoszmajsak: Just wondering what stands behind the selection of these values? (initial delay and period)

bartoszmajsak: In other words why those numbers? :)

pkg/plugin/probes_handler.go
bartoszmajsak: @hemanik this could help https://blog.questionable.services/article/testing-http-handlers-go/

pkg/plugin/probes-handler/probes_handler.go
```diff
@@ -0,0 +1,18 @@
...
+	"log"
+	"net/http"
+	"os"
+)
+
+// NewProbesHandler registers liveliness and readinesss probes for /version endpoint
+func NewProbesHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		_, err := w.Write([]byte(os.Getenv("VERSION")))
```
dipak-pawar: @hemanik How about returning `version: 406ea76-1524559762` instead of only `406ea76-1524559762`?

bartoszmajsak: Can we respond using proper json, content type and so on?

pkg/plugin/probes-handler/probes_handler.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// NewProbesHandler registers liveliness and readinesss probes for /version endpoint
+func NewProbesHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		w.Header().Set("Content-Type", "application/json")
+		_, err := w.Write(getJSONContent())
+		if err != nil {
+			log.Printf("Writing failed: %v", err)
```
MatousJobanek: the logger that is used here is something different than the one that is used in the rest of the project. Please, use the same one. 
For more information see:
https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/log/log.go
https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/plugin/plugin_init.go#L105
https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/plugin/plugin_init.go#L81

in addition, when there is an error in the probe, then the level should be higher - I would use `error` for it

pkg/plugin/probes-handler/probes_handler.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func getJSONContent() []byte {
+	version, found := os.LookupEnv("VERSION")
+	if !found {
+		version = "UNKNOWN"
+	}
+	js, err := json.Marshal(Probe{version})
+	if err != nil {
+		log.Printf("Marshalling failed: %v", err)
```
MatousJobanek: instead of just logging the error should be returned and checked in the handler function

pkg/plugin/probes-handler/probes_handler.go
```diff
@@ -0,0 +1,37 @@
...
+		}
+	})
+}
+
+func getJSONContent() []byte {
+	version, found := os.LookupEnv("VERSION")
+	if !found {
+		version = "UNKNOWN"
+	}
+	js, err := json.Marshal(Probe{version})
```
MatousJobanek: it's just a detail - in the rest of the project, when a constructor is used for instantiation, then all the keys are specified.

pkg/plugin/probes-handler/probes_handler_test.go
```diff
@@ -0,0 +1,64 @@
...
+		probesEndpoint = "/version"
+		version        = "xxxxxxxx-xxxxxxxxxx"
+	)
+
+	var _ = BeforeSuite(func() {
+		os.Setenv("VERSION", version)
+	})
+
+	var _ = AfterSuite(func() {
+		os.Clearenv()
```
MatousJobanek: `Clearenv` deletes all environment variables, which is something that we don't want to do. There should be probably used `Unsetenv`. Or, which would be cool, at the line 22 get the environment value that is already set and here reset it back.

pkg/plugin/probes-handler/probes_handler_test.go
```diff
@@ -0,0 +1,64 @@
...
+				w.WriteHeader(http.StatusBadGateway)
+			}
+			request := httptest.NewRequest("GET", probesEndpoint, nil)
+			response := httptest.NewRecorder()
+
+			// when
+			handler(response, request)
+
+			// then
+			Expect(response.Body.String()).To(Equal(""))
```
MatousJobanek: if I understand it correctly, when it fails/is unhealthy then it returns an empty string. is this really expected? what was behind this decision?
looking here https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/
there is written
> Any code greater than or equal to 200 and less than 400 indicates success. Any other code indicates failure.

this would mean that even when the proble is broken down then it would return 200.

pkg/plugin/probes-handler/probes_handler_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+			// then
+			actualBody := probeshandler.Probe{}
+			json.Unmarshal(response.Body.Bytes(), &actualBody)
+			Expect(actualBody).To(Equal(expectedBody))
+		})
+	})
+
+	Context("When in unhealthy state", func() {
+		It("should return empty response body", func() {
```
MatousJobanek: btw, does anything from this test verify anything from the probe functionality?

hemanik: My intention with this test was to test for the unhealthy state. The test does not verify the probe functionality directly but what I tried testing here was the end-point.

pkg/plugin/probes-handler/probes_handler_test.go
hemanik: Here, my intention was to forcefully fail and check for the unhealthy state. 
> Any code greater than or equal to 200 and less than 400 indicates success. Any other code indicates failure.

As mentioned here, I add a dummy handler that returns the error code 502 that exceeds the success code for the probe forcing it to fail as can be seen in the output when tested manually.

```
Events:
  FirstSeen	LastSeen	Count	From			SubObjectPath			Type		Reason			Message
  ---------	--------	-----	----			-------------			--------	------			-------
  1m		1m		1	default-scheduler					Normal		Scheduled		Successfully assigned test-keeper-4-ggj4z to localhost
  1m		1m		1	kubelet, localhost					Normal		SuccessfulMountVolume	MountVolume.SetUp succeeded for volume "plugins" 
  1m		1m		1	kubelet, localhost					Normal		SuccessfulMountVolume	MountVolume.SetUp succeeded for volume "oauth" 
  1m		1m		1	kubelet, localhost					Normal		SuccessfulMountVolume	MountVolume.SetUp succeeded for volume "sentry-dsn" 
  1m		1m		1	kubelet, localhost					Normal		SuccessfulMountVolume	MountVolume.SetUp succeeded for volume "hmac" 
  1m		1m		1	kubelet, localhost					Normal		SuccessfulMountVolume	MountVolume.SetUp succeeded for volume "default-token-h8t5s" 
  1m		1m		1	kubelet, localhost	spec.containers{test-keeper}	Normal		Pulling			pulling image "172.30.1.1:5000/myproject/test-keeper:6d586fe-1524764442"
  1m		1m		1	kubelet, localhost	spec.containers{test-keeper}	Normal		Pulled			Successfully pulled image "172.30.1.1:5000/myproject/test-keeper:6d586fe-1524764442"
  1m		1m		1	kubelet, localhost	spec.containers{test-keeper}	Normal		Created			Created container
  1m		1m		1	kubelet, localhost	spec.containers{test-keeper}	Normal		Started			Started container
  1m		1m		3	kubelet, localhost	spec.containers{test-keeper}	Warning		Unhealthy		Liveness probe failed: HTTP probe failed with statuscode: 502
  1m		9s		11	kubelet, localhost	spec.containers{test-keeper}	Warning		Unhealthy		Readiness probe failed: HTTP probe failed with statuscode: 502
``` 
In this case, hitting the `/version` end-point returns empty value. Covering this case with the tests was my intention. 
Not sure, if this was the best approach to test it and I can update it if you can suggest any better way.

pkg/plugin/probes-handler/probes_handler_test.go
hemanik: The variable I am trying to access is exposed inside the docker container and is not exposed in the test scope.
>  which would be cool, at the line 22 get the environment value that is already set and here reset it back.

So, the environment variables return empty values and the only variable set here is the one that is explicitly set in the `BeforeSuite` and this is what is cleared up in the `AfterSuite`.

MatousJobanek: > The variable I am trying to access is exposed inside the docker container and is not exposed in the test scope.

when the test is executed then the variable set on line 22 is exposed in the test scope. Then on line 26 all variables  (which means everything that is locally set) are deleted

> So, the environment variables return empty values and the only variable set here is the one that is explicitly set in the BeforeSuite and this is what is cleared up in the AfterSuite.

if I have set the variable `VERSION` for any reason locally, then this test would reset my original value.

pkg/plugin/probes-handler/probes_handler.go
```diff
@@ -13,25 +15,36 @@ type Probe struct {
 }
 
 // NewProbesHandler registers liveliness and readinesss probes for /version endpoint
-func NewProbesHandler() http.Handler {
+func NewProbesHandler(log *logrus.Entry) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+		serverError := func(action string, err error) {
+			log.WithError(err).Errorf("Error %s.", action)
```
MatousJobanek: Maybe some additional context in the message would help. eg: `Probe handler failed when`

pkg/plugin/probes-handler/probes_handler.go
```diff
@@ -13,25 +15,36 @@ type Probe struct {
...
+			msg := fmt.Sprintf("500 Internal server error %s: %v", action, err)
```
MatousJobanek: I would use the constant from the `http` package instead

pkg/plugin/probes-handler/probes_handler_test.go
```diff
@@ -29,7 +30,8 @@ var _ = Describe("Test liveliness and readiness probes.", func() {
 	Context("When in healthy state", func() {
 		It("should return plugin version in response body", func() {
 			// given
-			probesHandler := probeshandler.NewProbesHandler()
+			var log *logrus.Entry
```
MatousJobanek: for the test purposes, the `NewTestLogger()` function should be used

pkg/plugin/probes-handler/probes_handler_test.go
hemanik: Removing the test as discussed offline.

pkg/plugin/probes-handler/probes_handler_test.go
hemanik: Resetting the value to the original after suite as discussed offline.

Issue #100, opened 2018-04-10
Join saas-errortracking repository


bartoszmajsak:
All services deployed as part of OSIO (including Sentry) have their "deployment" repositories. We will have to create one following the process of [`saasherder`](https://github.com/openshiftio/saasherder)

**UPDATE**
Based on the conversation with @aslakknutsen and  @kbsingh we should make our new home part of "errortracking" project, so we need to move our templates to https://github.com/openshiftio/saas-errortracking repository.

MatousJobanek:
Before I do the PR to the `saas-errortracking` repo I need to finish one task:

- [ ] tag the images so we can support the "latest" tag 

and clarify one thing:
* the registry we want to push/pull the images to/from. There are two options:
  * `push.registry.devshift.net` - supported by centos-ci environment
    * the deployment could be automated by centos-ci deploy job (credentials are stored there)
  * `docker.io`
    * to make it automated by centos-ci we would need to somehow provide the credentials there. Or do it manually

based on the decision, I'll need to potentionally also add the `cico-deploy.sh` script to our project and send a PR to `openshiftio-cico-jobs` repo

bartoszmajsak:
I am not sure if we need `latest` - it's a moving target and we don't know what exact version is running on prod until we extract it from somewhere. Unless it's needed by the cico automation process I would stick with version tags (as we do now).

bartoszmajsak:
As for the registry I think we should be pushing to devshift, especially when using centos and soon RHEL.

Docker hub could be nice for streamlined upstream images, so I think we will come back to the discussion about pushing there later on anyway.

MatousJobanek:
as for the latest. there are two reasons to do that:
1. it is one of the "required" steps for the cico build process
2. we would need to change the parameter in `saas-errortracking` config file for every update of the docker image (change of the version tag).

as for the registry - yeah, I was kind of expecting that the devshift is the correct one, but wasn't 100% sure

bartoszmajsak:
Ok, then the process needs that. We have version info exposed anyway and saasherder tracks it by commit hashes, so I think we are good.

MatousJobanek:
Both the devshift thingy and the latest tags are part of: https://github.com/arquillian/ike-prow-plugins/pull/132

MatousJobanek:
[...]

PR #132, opened 2018-04-26
chore(#97): adds centos-ci script that deploys images to devshift

MatousJobanek:
* added a new cico script `cico_build_deploy.sh` that deploys images to devshift
* add new targets in `Makefile.deploy.prow` to encapsulate actions that compile binary, build images and push to the registry (without processing the template)
* every image version is tagged - adds the possibility of using `latest` tag
* `latest` tag is used as default one in templates
* fixed dockerfile - added missing environments
* extracted logic that retrieves prow version to separated script so I can use it in the cico script as well

Fixes: #97

### DIFF
--- a/cico_setup.sh
+++ b/cico_setup.sh
@@ -59,6 +59,7 @@ function deploy() {
 
   # compile, build and deploy the hook
   export PROW_VERSION=`./prow_version.sh | cut -c1-${DEVSHIFT_TAG_LEN}`
+  make docker-build-hook
   make deploy-hook
 
   # compile, build and deploy plugins
--- a/cico_setup.sh
+++ b/cico_setup.sh
@@ -30,7 +30,7 @@ function install_deps() {
 }
 
 function run_build() {
-  make build
+  make docker-build
   echo "CICO: ran build"
 }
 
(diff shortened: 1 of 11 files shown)


cluster/hook-template.yaml
```diff
@@ -9,6 +9,7 @@ parameters:
   value: arquillian
 - name: VERSION
   required: true
+  value: latest
```
dipak-pawar: How does the `latest` value help here? What's wrong with particular version?

bartoszmajsak: Reading issue comment can shed some light here. Plus this discussion I believe https://github.com/arquillian/ike-prow-plugins/issues/100#issuecomment-384276306

dipak-pawar: thanks for pointers.

cico_setup.sh
```diff
@@ -30,7 +30,7 @@ function install_deps() {
 }
 
 function run_build() {
-  make build
+  make docker-build
```
dipak-pawar: @MatousJobanek I can't see `make docker-build` target .

alien-ike:
Hey @dipak-pawar! It seems you tried to trigger `/ok-without-tests` command, but this will not have any effect due to insufficient permission. You have to be admin or requested reviewer or pull request approver, but not pull request creator for this command to take an effect.

cico_setup.sh
MatousJobanek: see here https://github.com/arquillian/ike-prow-plugins/blob/master/.make/Makefile.docker.build#L64-L68

MatousJobanek:
Now, I have realized that I have a bug there - I'm calling `deploy-hook` and `deploy-plugins` outside of the docker container, but there are not installed the go binaries. I'll fix it

MatousJobanek:
@dipak-pawar as I had to merge the changes from master, could you please re-mark it as "ok-without-tests" Thanks

MatousJobanek:
[test]

MatousJobanek:
[test]

Issue #101, opened 2018-04-10
Metrics collection


bartoszmajsak:
### Infrastructure metrics

- goroutines
- gc pause
- gc count
- memory allocations
- memory frees
- ....

### Service metrics

- rate limits
- incoming webhooks
- handled events
- ....

#### Test keeper
- /ok-without-tests vs size of the PR
- PRs with and without tests
- ....

bartoszmajsak:
### PCP

https://github.com/fabric8-services/fabric8-notification/blob/2358e9e9cbf9be6728a5a08106533c35dd16b8f2/pcp.repo

https://github.com/fabric8-services/fabric8-notification/blob/2358e9e9cbf9be6728a5a08106533c35dd16b8f2/notification%2Bpmcd.sh

### Prometheus

Adding Prometheus to our code could be done in a similar fashion as in other OSIO golang services:

https://github.com/fabric8-services/fabric8-notification/blob/2358e9e9cbf9be6728a5a08106533c35dd16b8f2/main.go#L149-L162

then in PCP config

https://github.com/fabric8-services/fabric8-notification/blob/2358e9e9cbf9be6728a5a08106533c35dd16b8f2/notification%2Bpmcd.sh#L38

bartoszmajsak:
### Custom metrics

https://github.com/fabric8-services/fabric8-wit/blob/master/metric/

bartoszmajsak:
Lets use this issue to also discuss other metrics we would like to gather.

dipak-pawar:
Next Steps:
 **Service Metrics**
- [ ] rate limits
- [ ] incoming webhooks
- [ ] handled events

**Plugin metrics**
- Test Keeper
 - [ ] /ok-without-tests vs size of the PR
 - [ ] PRs with and without tests

dipak-pawar:
@bartoszmajsak I don't see how can I implement  x `vs` y graph in prometheus metrics [types](https://prometheus.io/docs/concepts/metric_types/). For Test Keeper metrics (/ok-without-tests vs size of the PR, PRs with and without tests) . Do you want me to implement it as counter/gauge/histogram metrics?

I hope I am not missing anything?

MatousJobanek:
Hi @dipak-pawar,
if I'm not mistaken, the main point in test-keeper metrics was to measure the quality of the PRs. How many times was provided a PR with/without tests, so the `/ok-without-tests` command had to be used.
[...]

PR #133, opened 2018-04-26
chore(pcp/prometheus): enable collection of pod data for monitoring

dipak-pawar:
This PR adds pcp/prometheus to collect intra-pod data for monitoring.

#### Changes proposed in this PR
* Adds pcp and prometheus configuration
* Creating non-root user `prow`
* Adds `/metrices` endpoint


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/arquillian/ike-prow-plugins/dfd7d4a075c0de63a88af8780ef38c7fd52ba7a4/test-keeper.yaml).

Dockerfile.prow
```diff
@@ -2,12 +2,33 @@ FROM centos:7
...
+# would prefer only pmcd, and not the /bin/pm*tools etc.
+COPY pcp.repo /etc/yum.repos.d/pcp.repo
+RUN yum install -y pcp pcp-pmda-prometheus && yum clean all && \
+    mkdir -p /etc/pcp /var/run/pcp /var/lib/pcp /var/log/pcp  && \
+    chgrp -R root /etc/pcp /var/run/pcp /var/lib/pcp /var/log/pcp && \
+    chmod -R g+rwX /etc/pcp /var/run/pcp /var/lib/pcp /var/log/pcp
+COPY ./prow+pmcd.sh /prow+pmcd.sh
+EXPOSE 44321
+
+COPY bin/${BINARY} ${INSTALL_PREFIX}/binary
```
MatousJobanek: In all other OSIO projects, there is another directory `bin` in the path the binary is copied to. I would keep it consistent - so use  `${INSTALL_PREFIX}/bin/binary`. But that is only my opinion.

dipak-pawar:
@MatousJobanek
> you have created a separated docker image for hook and for plugins (named .prow but when it is separated then it should be probably .plugin). Looking at the hook implementation they listen to the /metrics endpoint using the Prometheus handler as well. Is there some reason why we cannot use the endpoint?

We have `/metrics` endpoint for each plugin. Just that I am not installing  `pcp,pcp-pmda-prometheus`  in hook container, because I am not sure do we really need to monitor `hook`  

> I'm missing any metric in this PR. Does it mean that it is still work-in-progress?

I am planning to add it as a part of another PR.

MatousJobanek:
> We have /metrics endpoint for each plugin. Just that I am not installing pcp,pcp-pmda-prometheus in hook container, because I am not sure do we really need to monitor hook

I would do it if it is possible - to monitor the hook as well.

> I am planning to add it as a part of another PR.

then please create corresponding issues for all PRs and reference them here https://github.com/arquillian/ike-prow-plugins/issues/101

Issue #70, opened 2018-03-27
Make sure the bypass for missing tests is retained when the PR is updated


xcoulon:
For now, each time a PR is updated, the `/ok-without-tests` comment needs to be input again, otherwise the `alien-ike/test-keeper` job will complain again about the lack of tests (which, in some edge case, is acceptable)

bartoszmajsak:
The solution could be to check if there is a `/ok-without-tests` comment made by any of the authorized users.

PR #134, opened 2018-04-27
fix(#70): test-keeper bypass status stays once is set

MatousJobanek:
* when the test-keeper bypass command is used, then the status stays even when the PR is updated
  * this means that it sends the bypass status for the updated hash with the same context (same `@sender`)
* adds `IssueCommentsLazyLoader` for loading issue comments only once

Fixes: #70


MatousJobanek:
[test]

pkg/plugin/test-keeper/event_handler.go
```diff
@@ -67,16 +67,40 @@ func (gh *GitHubTestEventsHandler) handlePrEvent(log log.Logger, event *gogh.Pul
 	if !utils.Contains(handledPrActions, *event.Action) {
 		return nil
 	}
+	commentsLoader := ghservice.NewIssueCommentsLazyLoader(gh.Client, event.PullRequest)
 
-	return gh.checkTestsAndSetStatus(log, event.PullRequest)
+	bypassed, user := gh.checkIfBypassed(log, commentsLoader, event.PullRequest)
+	if bypassed {
```
dipak-pawar: I believe we should check if recently pushed changes have some tests, if yes then set correct status as tests found and pass this check if not found then bypass this depending on `/ok-without-tests`.

This way we'll set correct status with a message. Currently, we are not handling above mentioned scenario

MatousJobanek: Good point @dipak-pawar. The correct message is something definitely helpful. 
On the other hand, the build will be always green - no matter if any new test has been added or removed, only the message could change. Not sure if that is a good reason for using resources and requests. In addition, if the message stays same: `PR is fine without tests says @...` it says the truth if the comment is there. If anyone wants to re-execute the plugin without the bypass, it is still possible to remove the comment.

dipak-pawar: I think we should make our plugin as smart as possible, instead of asking a user to delete the comment to execute test check. We can still say `looks like PR is fine without tests says @..., but found some tests` and make check green. I would focus on making it smart instead of caring for using resources and requests.

I'll leave it to you. :smile:

MatousJobanek: Although I'm not really convinced that it brings any additional value, I moved the check so it checks the presence of the test files every time. Please take a look at the very last commit

Issue #135, opened 2018-04-27
Remove unnecessary logging of GH API errors


MatousJobanek:
when I was running tests with logging set on, I've noticed that there is a lot of error logs visible in my console. All of them were referencing to nill error:
`{"level":"error","msg":"failed while trying to call GitHub API %!s(\u003cnil\u003e).","time":"2018-04-27T11:35:43+02:00"}`
That should be removed since any potential "final" error is logged in the `retry_client.go`

PR #136, opened 2018-04-27
fix(#135): removes unnecessary logging of GH API errors

MatousJobanek:
Fixes: #135

### DIFF
--- a/pkg/github/client/client.go
+++ b/pkg/github/client/client.go
@@ -7,6 +7,7 @@ import (
 	"github.com/arquillian/ike-prow-plugins/pkg/scm"
 	gogh "github.com/google/go-github/github"
 	"golang.org/x/oauth2"
+	"fmt"
 )
 
 type client struct {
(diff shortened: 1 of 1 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/arquillian/ike-prow-plugins/da9fc5b0369eae9a4508b1063a98e861ad15fe0f/test-keeper.yaml).

MatousJobanek:
Good point @bartoszmajsak, that is another "bug" that I missed :-).
The whole if should be:
```go
	if e == nil && response != nil && response.StatusCode >= 404 {
		return fmt.Errorf("server responded with %d status", response.StatusCode)
	}
```
The point is that we need to create a new error in case the retrieved is nil but status code is >= 404  - to make the retry function working. 
In addition, I guess that  we don't want to log every error, but only the aggregated one in [retry_client.go](
https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/github/client/retry_client.go#L34-L38)
This takes me to the fact that the function name should be different. eg: `checkHTTPCode` or `handleHTTPError` or something like that

MatousJobanek:
@hemanik @dipak-pawar could you please take a look at this PR when you find some time?

alien-ike:
Hey @dipak-pawar! It seems you tried to trigger `/ok-without-tests` command, but this will not have any effect due to insufficient permission. You have to be admin or requested reviewer or pull request approver, but not pull request creator for this command to take an effect.

PR #142, opened 2018-04-30
fix(#141): add service metrics for monitoring

### LINKED ISSUE #141: Add Service Metrics for monitoring
- rate limits
- incoming webhooks
- handled events

dipak-pawar:
Fixes #141 

**Metrics output on `/metrics` endpoint**
```bash
# HELP github_rate_limits Rate limit for GitHub API calls
# TYPE github_rate_limits gauge
github_rate_limits{type="core"} 4984
github_rate_limits{type="search"} 30
# HELP handled_events_counter A counter of handled events.
# TYPE handled_events_counter counter
handled_events_counter{event_type="pull_request"} 3

# HELP webhook_counter A counter of the webhooks made to service.
# TYPE webhook_counter counter
webhook_counter{full_name="dipak-pawar/ike-prow-plugin-test"} 2
webhook_counter{full_name="dipak-pawar/ike-prow-plugins"} 1
```


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/arquillian/ike-prow-plugins/873e30989adc358dfae3d8beb7c3e1edce62eeb9/test-keeper.yaml).

pkg/server/metrics_test.go
```diff
@@ -0,0 +1,110 @@
...
+			logger.Errorf("Error sending hook: %v", err)
+		}
+
+		defer s.Close()
+	})
+
+	It("should count incoming webhook", func() {
+		// given
+		eventType := "pull_request"
+		expected := 1
```
MatousJobanek: is this really necessary - to use it as variables?

pkg/server/metrics_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+		// then
+		Ω(err).ShouldNot(HaveOccurred())
+		Expect(count(counter)).To(Equal(expected))
+	})
+
+	It("should count handled events", func() {
+		// given
+		eventType := "pull_request"
+		expected := 2
```
MatousJobanek: correct me if I'm wrong - this test depends on the previous one, right? it should be possible to run the test separately and in any order.

pkg/server/server.go
```diff
@@ -49,6 +50,12 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		},
 	)
 
+	if counter, err := s.Metrics.WebhookCounter.GetMetricWithLabelValues(eventType); err != nil {
```
MatousJobanek: what is the difference in counting webhook requests with eventType payload and counting GH events with the same eventType payload on the line 70? isn't it same?

pkg/server/server.go
```diff
@@ -59,6 +66,10 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		})
 	}
 
+	if eventType == string(github.PullRequest) || eventType == string(github.IssueComment) {
```
MatousJobanek: shouldn't that be already filtered via plugins configuration in `plugins.yaml` file?

pkg/server/metrics_test.go
```diff
@@ -0,0 +1,110 @@
...
+			Reply(200).
+			Body(test.FromFile("../github/client/test_fixtures/gh/low_rate_limit.json"))
+
+		gock.New("https://api.github.com").
+			Get("/repos/owner/repo/pulls/123/files").
+			Reply(200).
+			BodyString("[]")
+
+		client := ghclient.NewRateLimitWatcherClient(test.NewDefaultGitHubClient(), logger, 10)
+		client.ListPullRequestFiles("owner", "repo", 123)
```
MatousJobanek: correct me if I'm wrong - this is the `// when` part - when the `client.ListPullRequestFiles` is called then it should gather the rate limit metrics

pkg/server/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+		Help: "A counter of the webhooks made to service.",
+	}, []string{"event_type"})
+
+	handledEventsCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "handled_events_counter",
+		Help: "A counter of handled events.",
+	}, []string{"event_type"})
+)
+
+func init() {
```
MatousJobanek: instead of initializing it immediately name it for example `registerMetrics` and call from the `plugin_init.go`

pkg/server/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+// Metrics is a set of metrics gathered by the Ike Prow Plugin.
+// It includes rate limit, incoming webhooks. handled events.
+type Metrics struct {
+	RateLimit            *prometheus.GaugeVec
+	WebhookCounter       *prometheus.CounterVec
+	HandledEventsCounter *prometheus.CounterVec
+}
+
+// NewMetrics creates a new set of metrics for the Ike Prow Plugin.
+func NewMetrics() *Metrics {
```
MatousJobanek: is it necessary when it is used only in test and only at one place?

pkg/server/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+	}, []string{"event_type"})
+
+	handledEventsCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "handled_events_counter",
+		Help: "A counter of handled events.",
+	}, []string{"event_type"})
+)
+
+func init() {
+	prometheus.MustRegister(rateLimit)
```
MatousJobanek: please use `prometheus.register` and handle the error properly

dipak-pawar: @MatousJobanek I believe in an ideal scenario in any service, Prometheus should register all required collector. but if something bad happens shouldn't be it panic?

In face Kubernetes project is also using this same way https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/metrics/metrics.go#L45

pkg/server/metrics.go
dipak-pawar: now using it two places. :)

pkg/server/server.go
dipak-pawar: sorry, my bad. I have improved in https://github.com/arquillian/ike-prow-plugins/pull/142/commits/958008f0b4e9a8f83c6b002eb8fdba394a8878ba

pkg/log/log.go
```diff
@@ -17,6 +17,7 @@ type Logger interface {
 	Warnf(format string, args ...interface{})
 	Error(args ...interface{})
 	Errorf(format string, args ...interface{})
+	Panic(args ...interface{})
```
MatousJobanek: Please, don't add such methods if not really necessary. see my comment in `metrics.go` file

pkg/plugin/plugin_init.go
```diff
@@ -94,7 +94,10 @@ func InitPlugin(pluginName string, newEventHandler EventHandlerCreator, newServe
 		100)
 
 	handler := newEventHandler(githubClient, *pluginBotName)
+
+	server.RegisterMetrics(logger)
 	pluginServer := newServer(webhookSecret, handler)
+	pluginServer.Metrics = server.NewMetrics()
```
MatousJobanek: this looks a bit weird - that it first registers the metrics and then calls the constructor. I think that this should be part of one step (method)

pkg/server/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+	HandledEventsCounter *prometheus.CounterVec
+}
+
+// NewMetrics creates a new set of metrics for the Ike Prow Plugin.
+func NewMetrics() *Metrics {
+	return &Metrics{
+		RateLimit:            rateLimit,
+		WebhookCounter:       webhookCounter,
+		HandledEventsCounter: handledEventsCounter,
+	}
```
MatousJobanek: if you want to use the instance of the `Metrics` struct as a reference to the variables (`rateLimit`, `webhookCounter` and `handledEventsCounter`) then it is not necessary to keep it as a global variable inside of this file, but only as the variables stored within the `Metrics` instance. Then the Report/Record functions could be used as methods with the receiver.

pkg/server/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func register(c prometheus.Collector, name string, log log.Logger) prometheus.Collector {
+	err := prometheus.Register(c)
+	if err != nil {
+		if are, ok := err.(prometheus.AlreadyRegisteredError); ok {
+			return are.ExistingCollector
+		}
+
+		log.Panic(map[string]interface{}{
```
MatousJobanek: you have three options to avoid adding the `Panic` method to the interface:
1. return an array of errors that should be reported inside of the `plugin_init.go` using `logger.WithError(err).Fatalf`. if the array is empty, then everything is OK
2. instead of using our interface as a type of the `log` parameter in this function use the `*logrus.Entry` directly so you can fail using the same method  `logger.WithError(err).Fatalf` directly from this place (don't forget to accumulate the error)
3. keep the type of the parameter as it is (our interface), log every "register error" separately using `error` method. When there is some error return a new one (eg. `errors.New("Prometheus metrics registration failed. See preceding errors.")errors.New("Prometheus metrics registration failed. See preceding errors.")`) to the `plugin_init.go` where you can fail.

pkg/server/server.go
```diff
@@ -59,8 +60,28 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+func recordIncomingWebHooks(s *Server, l log.Logger, label string) {
+	if counter, err := s.Metrics.WebhookCounter.GetMetricWithLabelValues(label); err != nil {
+		l.Errorf("Failed to get metric for Repository: %q. Cause: %q", label, err)
+	} else {
+		counter.Inc()
+	}
+}
+
+func recordHandledEvents(s *Server, l log.Logger, label string) {
```
MatousJobanek: both of the `record*` methods could be moved to `metrics.go`

pkg/github/client/rate_limit_watcher_client.go
```diff
@@ -16,24 +16,37 @@ type rateLimitWatcher struct {
...
 }
 
 func (r rateLimitWatcher) logRateLimitsAfter(f func()) {
 	f()
 	r.logRateLimits()
 }
 
-func (r rateLimitWatcher) logRateLimits() {
+// GetRateLimits retrieves the rate limits for the current GH client
+func GetRateLimits(c Client) *gogh.RateLimits {
```
MatousJobanek: I'm slightly confused here - what was wrong with the previous solution https://github.com/arquillian/ike-prow-plugins/pull/142/commits/c499fdf0e20309722e9154e5304bdd61222e0305#diff-535bf210f0256dc391d9eff7602869f5L36

pkg/plugin/plugin_init.go
```diff
@@ -94,8 +96,17 @@ func InitPlugin(pluginName string, newEventHandler EventHandlerCreator, newServe
...
+
 	pluginServer := newServer(webhookSecret, handler)
 
+	if errs := pluginServer.RegisterMetrics(); len(errs) > 0 {
+		msg := ""
+		for index, e := range errs {
+			msg = msg + fmt.Sprintf("\n%d. [%s]", index+1, e.Error())
+		}
+		logger.WithError(errors.New(msg)).Fatal("Prometheus metrics registration failed!!!")
+	}
```
MatousJobanek: I was thinking about something like this:
```go
	if errs := pluginServer.RegisterMetrics(); len(errs) > 0 {
		errLog := logger
		for _, e := range errs {
			errLog = errLog.WithError(e)
		}
		errLog.Fatal("Prometheus metrics registration failed!")
	}
```
could that work too?

pkg/plugin/pr-sanitizer/cmd/main.go
```diff
@@ -42,6 +42,11 @@ func (gh *GitHubLabelsEventsHandler) HandleEvent(log log.Logger, eventType githu
 	return nil
 }
 
+// GhClient created while event handler initialization which is used to communicate with the GitHub API
+func (gh *GitHubLabelsEventsHandler) GhClient() ghclient.Client {
+	return gh.Client
+}
```
MatousJobanek: this is Java way of doing it. We should use a different approach to solve the problem

pkg/server/metrics.go
```diff
@@ -0,0 +1,96 @@
+package server
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/arquillian/ike-prow-plugins/pkg/log"
+	"github.com/arquillian/ike-prow-plugins/pkg/github/client"
+)
+
+// RegisterMetrics registers prometheus collectors to collect metrics
+func (s *Server) RegisterMetrics() []error {
```
MatousJobanek: what is the receiver good for? there is nothing used here from the struct. It would make more sense to return an instance of the `Metrics` and then assign it in the caller...

pkg/server/metrics.go
```diff
@@ -0,0 +1,96 @@
...
+	} else {
+		errors = append(errors, e)
+	}
+
+	if len(errors) > 0 {
+		return errors
+	}
+
+	s.Metrics = metrics
+	return make([]error, 0);
```
MatousJobanek: you should be able to return the original `errors` variable without checking its content - that is checked in the caller

pkg/server/metrics.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/arquillian/ike-prow-plugins/pkg/log"
+	"github.com/arquillian/ike-prow-plugins/pkg/github/client"
+)
+
+// RegisterMetrics registers prometheus collectors to collect metrics
+func (s *Server) RegisterMetrics() []error {
+	errors := make([]error, 0, 3)
+	metrics := NewMetrics()
+	if rateLimitCollector, e := registerOrGet(metrics.RateLimit); e == nil {
```
MatousJobanek: I think that it could be shorter: 
```go
metrics.RateLimit = register(metrics.RateLimit, &errors).(*prometheus.GaugeVec)
```

MatousJobanek:
[test]

MatousJobanek:
[test]

dipak-pawar:
[test]

Issue #28, opened 2018-03-06
WIP prefix configurable


bartoszmajsak:
Add ability to define custom prefix and label to be used in simple `yaml` file. For the moment we support `WIP`, but people tend to use `[WIP]`, `[WIP]:`, `WIP:`, `DO-NOT-MERGE` etc. We should make it configurable.

MatousJobanek:
I would add the similar prefixes `[WIP]`, `[WIP]:`, `WIP:`, `(WIP)`, `(WIP):` to the default set - match them automatically. But having it configurable is a must!

bartoszmajsak:
Yes, defaults need to be there as we don't want ppl to configure mostly the same things over and over again. So I can think of

- WIP
- DO NOT MERGE
- DON'T MERGE
- WORK-IN-PROGRESS

with all the flavours you listed above (with `:` and `[ ]`, `( )`)

They should be supported by default.

PR #143, opened 2018-05-04
 feat(#28): makes wip prefix configurable.

hemanik:
This PR brings in the following features to the WIP plugin.

- Adds some common, work in progress prefixes (WIP, DO NOT MERGE, DON'T MERGE, WORK-IN-PROGRESS case insensitive with combinations of `( )`, `[ ]`, `:`) as default to be added to the PR title. 
- Allow for user defined prefixes by making it configurable by setting the `title_prefixes` property in the configuration file for the work in progress plugin.  

Fixes: #28


MatousJobanek:
- [ ] Please add also the description to the documentation

pkg/plugin/work-in-progress/event_handler_gh_test.go
```diff
@@ -80,8 +84,31 @@ var _ = Describe("Test Keeper Plugin features", func() {
 			Ω(err).ShouldNot(HaveOccurred())
 		})
 
+		It("should mark opened PR as work-in-progress when prefixed with configured prefix", func() {
```
MatousJobanek: `should mark opened PR as work-in-progress when title starts with configured prefix`

pkg/plugin/work-in-progress/configuration_gh_test.go
```diff
@@ -0,0 +1,66 @@
...
+				Owner:    "owner",
+				RepoName: "repo",
+				Hash:     "46cb8fac44709e4ccaae97448c65e8f7320cfea7",
+			}
+
+			// when
+			configuration := wip.LoadConfiguration(logger, change)
+
+			// then
+			Expect(configuration.Prefix).To(ConsistOf("[work in progress]", "work in progress"))
```
MatousJobanek: please verify also that the `combine_defaults` is equal true

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -40,6 +38,7 @@ type GitHubWIPPRHandler struct {
 
 var (
 	handledPrActions = []string{"opened", "reopened", "edited"}
+	defaultPrefixes  = []string{"WIP", "DO NOT MERGE", "DON'T MERGE", "WORK-IN-PROGRESS"}
```
dipak-pawar: @hemanik Are we missing `WORK IN PROGRESS` here?

MatousJobanek: according to the requested changes in the [issue](https://github.com/arquillian/ike-prow-plugins/issues/28#issuecomment-382414676) no. According to common sense, we could miss it, but I would say wait for some time to get the idea which patterns (prefixes) are used in the projects. After all, it is still configurable

dipak-pawar: @MatousJobanek I totally understand your point of this is configurable. But I think `Work In Progress` is common pattern which you can see in general

MatousJobanek: Personally, I use only `WIP` or `do not merge`. But that is my opinion. 
If you are used to using this prefix as well, then feel free to open a PR with this change ;-)

Issue #119, opened 2018-04-17
New home for configuration files


bartoszmajsak:
To simplify configuration management both from our perspective and end-users we should move configuration to one place - `.ike-prow/`.

This change requires following adjustments

- [x] change default location of the plugin to `.ike-prow/plugin-name.[yml|yaml]`
- [x] similar for plugin hint (convention over configuration as discussed here https://github.com/arquillian/ike-prow-plugins/pull/84#discussion_r179571129)
- [x] add this folder to default exclusions https://github.com/arquillian/ike-prow-plugins/blob/69435eb3bbd5135ee2ceed6c7a36a7d8859f4e72/pkg/assets/config/test-keeper.yaml#L131

bartoszmajsak:
Apart of the repository which @gastaldi linked, here are the other ones which we should open PR to with the new location of the config:

- [fabric8-wit](https://github.com/fabric8-services/fabric8-wit/pull/2011)
- https://github.com/fabric8-services/fabric8-auth

Please also review if their settings are not overlapping with our extended defaults. In such case, we can maybe even get rid of the files in the repos (like we can in `auth`).

dipak-pawar:
@bartoszmajsak I have gone through [fabric8-auth](https://github.com/fabric8-services/fabric8-auth),  [fabric8-wit](https://github.com/fabric8-services/fabric8-wit), Looks like we can get rid of the file `test-keeper.yaml` in both the repositories as our extended defaults are covering all expressions which is present in both repositories.  I will submit a PR once this get merged.

PR #144, opened 2018-05-07
fix(#119): add new home for configuration files

dipak-pawar:
Fixes: #119


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/arquillian/ike-prow-plugins/cb93dba0df1a9db0c2751573dfc6fc83d5443488/test-keeper.yaml).

MatousJobanek:
Thanks @dipak-pawar for your work, unfortunately, I'm afraid that the task was misunderstood. The new default location of the config files `.ike-prow/` should be something that is automatically checked for the presence of the files - for both the configuration file and plugin hint file (with that fact that the location of the plugin hint file can be still customized in config file). It should work in the same way as the `.github` directory works for the GH templates.
The `pkg/assets/config/test-keeper.yaml` should stay there where it was - this contains default inclusions and exclusions applicable for all projects the plugins are used in. The `.ike-prow/test-keeper.yaml` file contains additional configuration specific for the particular project - our own one is here: https://github.com/arquillian/ike-prow-plugins/blob/master/test-keeper.yaml that one should have been moved to the directory.

MatousJobanek:
[test]

dipak-pawar:
@MatousJobanek can you look into this when you find some time?

pkg/assets/config/test-keeper.yaml
```diff
@@ -134,3 +135,4 @@ skip_validation_for:
 
   # Our own configurations
   - 'regex{{test-keeper\.y[a]?ml$}}'
```
MatousJobanek: is this line still necessary when the configs should be inside of the directory that is excluded?

.ike-prow/test-keeper.yaml
```diff
@@ -0,0 +1,2 @@
+skip_validation_for: 
+  - 'cluster/'
```
MatousJobanek: there is still also the original file https://github.com/dipak-pawar/ike-prow-plugins/blob/c32f52a822cedb073f21298dda3e98ecb424bf82/test-keeper.yaml - it should be removed (moved here)

pkg/github/service/config_loader.go
```diff
@@ -44,6 +44,16 @@ func (l *LoadableConfig) loadFromRawFile(pathTemplate string) config.Source {
...
 		l.BaseConfig.PluginName = l.PluginName
+
+		pluginHintPath := fmt.Sprintf(".ike-prow/%s_hint.md", l.PluginName)
+		hintURL := rawFileService.GetRawFileURL(pluginHintPath)
+
+		_, e := utils.GetFileFromURL(hintURL)
+
+		if e == nil {
+			l.BaseConfig.PluginHint = githubBaseURL + rawFileService.GetRelativePath(pluginHintPath)
+		}
```
MatousJobanek: this shouldn't be here. This function is for loading configuration file, not the *_hint file. 
In addition, on the line 51 you download the file - that should be lazy. Download it (and generate the default path) only when needed. 
This should be part of `comment_message.go`

pkg/plugin/test-keeper/configuration_gh_test.go
```diff
@@ -24,8 +24,10 @@ var _ = Describe("Test keeper config loader features", func() {
 
 		It("should load test-keeper configuration yml file", func() {
 			// given
+			NonExistingRawGitHubFiles(".ike-prow/test-keeper_hint.md")
```
MatousJobanek: I don't think that this is necessary as it checks the suffix

MatousJobanek:
- [ ] Please also change the documentation to explain the "workflow" of loading the `*_hint.md` files - that when nothing is set in config, then it is automatically loaded from the pre-defined location.

MatousJobanek:
@dipak-pawar  please, take a look at my last changes - if it makes sense to you and you are fine with it.

dipak-pawar:
@MatousJobanek Thanks for all changes. lgtm

MatousJobanek:
[test]

PR #146, opened 2018-05-15
Ignore codecov.io config file

xcoulon:
Configuration file can be `codecov.yml` or `.codecov.yml` as described in https://docs.codecov.io/docs/codecov-yaml#section-repository-yaml


xcoulon:
thanks, @dipak-pawar!

PR #149, opened 2018-05-17
chore(#150): adds RHEL based dockerfile used in CICO master build

MatousJobanek:
* added RHEL based dockerfile that should be pushed to devshift only from ci.centos.org master build
* renamed `Dockerfile.prow` to `Dockerfile.deploy` to keep the convention naming used in other OSIO projects

Fixes: #150


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/arquillian/ike-prow-plugins/cda15dd1b608e653bfd0339d158ad63ad3473323/test-keeper.yaml).

MatousJobanek:
I totally forgot to check the target - thanks a lot for pointing this out @jmelis and for the proposal

two execution of the script should pass without any problem:
* the docker image should be cleaned when the script stops
* even if it wasn't be cleaned up, the `prepare` function should pass twice without any problem

MatousJobanek:
[test]

Issue #160, opened 2018-05-23
WIP doesn't set status when a new commit is pushed


MatousJobanek:
WIP plugin should handle also `synchronize` PR action, otherwise, it doesn't set the status to the latest hash of the PR head.

PR #161, opened 2018-05-23
fix(#160): wip plugin handles also synchronize PR action

MatousJobanek:
Fixes: #160


alien-ike:
Hey @hemanik! It seems you tried to trigger `/ok-without-tests` command, but this will not have any effect due to insufficient permission. You have to be admin or requested reviewer or pull request approver, but not pull request creator for this command to take an effect.

dipak-pawar:
@MatousJobanek wouldn't it be good if we add test for this new synchronize action?

PR #168, opened 2018-05-28
fix(#166): uses gock networking only for POST requests

### LINKED ISSUE #166: Fix failing metrics test due to rate limit exceeds
Currently, metrics tests are failing due to exceeding rate limit. See https://travis-ci.org/arquillian/ike-prow-plugins/builds/384675732#L819

MatousJobanek:
The gock networking doesn't seem to work correctly when it is enabled for one whole test case. It should be enabled only for the particular request

Fixes #166

### DIFF
--- a/pkg/server/metrics_test.go
+++ b/pkg/server/metrics_test.go
@@ -57,11 +57,12 @@ var _ = Describe("Service Metrics", func() {
 		serverMetrics.RateLimit.Reset()
 		prometheus.Unregister(serverMetrics.HandledEventsCounter)
 		serverMetrics.HandledEventsCounter.Reset()
+		EnsureGockRequestsHaveBeenMatched()
 	})
-	AfterEach(EnsureGockRequestsHaveBeenMatched)
 
 	It("should count incoming webhook", func() {
 		// given
+		setGockMocks()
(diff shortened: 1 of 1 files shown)


pkg/server/metrics_test.go
```diff
@@ -119,6 +113,16 @@ var _ = Describe("Service Metrics", func() {
...
 })
 
+func setGockMocks() {
+	gock.New("https://api.github.com").
+		Get("/rate_limit").
+		Reply(200).
+		Body(FromFile("../github/client/test_fixtures/gh/low_rate_limit.json"))
+	gock.New("http://127.0.0.1").
+		Post("").
+		EnableNetworking()
```
dipak-pawar: Don't you think we need to `disableNetworking` after each tests or entire suite?

MatousJobanek: from the gock's code:
```go
// EnableNetworking enables the use real networking for the current mock.
func (r *Request) EnableNetworking() *Request {
	if r.Response != nil {
		r.Response.UseNetwork = true
	}
	return r
}
```
so I guess that it is tied to the particular request. Then calling the `defer gock.OffAll()` should be enough

Issue #6, opened 2018-02-26
PR title verification (semantic messages)


bartoszmajsak:
Verify if the title of the PR conforms with [semantic commit message](https://seesparkbox.com/foundry/semantic_commit_messages). If not, fail the check.

PR #169, opened 2018-05-29
feat(#6): PR Title Verification (Semantic Messages)

hemanik:
1. Introduces check for the PR Title Verification for the semantic message by identifying each PR into one of the categories - `chore`, `docs`, `feat`, `fix`, `refactor`, `style`, `test` or as per user defined configuration. 
2. Introduces configuration property `type_prefixes` for pr-sanitizer plugin to enable semantic message as per your organisation style (eg. [emoji-style types](https://medium.com/walmartlabs/semantic-commit-messages-with-emojis-dba2541cea9a) ) 

#### Needs update
- [x]  Documentation

 #### Fixes: #6


pkg/plugin/pr-sanitizer/test_fixtures/github_calls/custom_prefix_pr_opened.json
```diff
@@ -0,0 +1,459 @@
+{
+  "action": "edited",
```
MatousJobanek: the file has a name: `custom_prefix_pr_opened.json` but the action is `edited` - it shouldn't influence the logic anyhow, but it's good to keep it consistent

docs/fragments/pr-sanitizer.adoc
```diff
@@ -0,0 +1,51 @@
+== PR Sanitizer Plugin
+
+This plugin enforces that the PR title comforms with the semantic commit message style. Conformance with the semantic commit message style not only makes your changelog and git history clean and easy to navigate but also encourages making atomic commits that are more clearly understood.
```
dipak-pawar: How about adding  `inspired by` https://seesparkbox.com/foundry/semantic_commit_messages ?

pkg/plugin/pr-sanitizer/cmd/main.go
```diff
@@ -3,29 +3,18 @@ package main
 import (
 	"k8s.io/test-infra/prow/pluginhelp"
 
-	"github.com/arquillian/ike-prow-plugins/pkg/github"
 	"github.com/arquillian/ike-prow-plugins/pkg/github/client"
-	"github.com/arquillian/ike-prow-plugins/pkg/log"
 	pluginBootstrap "github.com/arquillian/ike-prow-plugins/pkg/plugin"
+	prsanitizer "github.com/arquillian/ike-prow-plugins/pkg/plugin/pr-sanitizer"
```
dipak-pawar: How about `prSanitizer` instead of  `prsanitizer`?

pkg/plugin/pr-sanitizer/configuration_gh_test.go
```diff
@@ -0,0 +1,66 @@
...
+			configuration := prsanitizer.LoadConfiguration(logger, change)
+
+			// then
+			Expect(configuration.TypePrefix).To(ConsistOf(":star:", ":package:", ":hammer_and_wrench:"))
+			Expect(configuration.Combine).To(Equal(true))
+		})
+
+		It("should not load pr-sanitizer configuration yaml file and return empty url when config is not accessible", func() {
+			// given
+			NonExistingRawGitHubFiles(".ike-prow/pr-sanitizer.yml", ".ike-prow/pr-sanitizer.yaml")
```
dipak-pawar: Only file name is sufficient `NonExistingRawGitHubFiles("pr-sanitizer.yml", "pr-sanitizer.yaml")`

pkg/plugin/pr-sanitizer/event_handler.go
```diff
@@ -0,0 +1,95 @@
...
+
+const (
+	// ProwPluginName is an external prow plugin name used to register this service
+	ProwPluginName = "pr-sanitizer"
+
+	// FailureMessage is a message used in GH Status as description when the PR title does not follow semantic message style
+	FailureMessage = "PR title does not conform with semantic commit message style."
+
+	// FailureDetailsLink is a link to an anchor in arq documentation that contains additional status details for failure state
+	FailureDetailsLink = "prsanitizer-failed"
```
dipak-pawar: How about `"pr-sanitizer-failed"`  to make it consistent with plugin  name?

pkg/plugin/pr-sanitizer/event_handler.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// SuccessMessage is a message used in GH Status as description when the PR title conforms to the semantic commit message style
+	SuccessMessage = "PR title conforms with semantic commit message style."
+	// SuccessDetailsLink is a link to an anchor in arq documentation that contains additional status details for success state
+	SuccessDetailsLink = "prsanitizer-success"
```
dipak-pawar: How about  `"pr-sanitizer-success"`  to make it consistent with plugin  name?

docs/fragments/pr-sanitizer.adoc
hemanik: Not sure if want to link to this? WDYT @MatousJobanek?

pkg/plugin/pr-sanitizer/cmd/main.go
hemanik: Removing it as it is not required at all.

pkg/plugin/pr-sanitizer/event_handler.go
```diff
@@ -0,0 +1,95 @@
...
+	if len(config.TypePrefix) != 0 {
+		if config.Combine {
+			prefixes = append(prefixes, config.TypePrefix...)
+		} else {
+			prefixes = config.TypePrefix
+		}
+	}
+
+	for _, prefix := range prefixes {
+		if strings.HasPrefix(strings.ToLower(title), strings.ToLower(prefix)) {
```
MatousJobanek: What if the PR is still "work in progress" and starts with prefix `WIP`? then this plugin would fail I guess, which is not good.

docs/fragments/pr-sanitizer.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+[source, shell, indent=0]
+----
+<type>(<scope>): <subject>
+----
+
+* where `<scope>` is optional.
+* and each of `<type>` categorizes the PR according to the change introduced (such as `feat` for introducing a new feature and `fix` for patching a bug in codebase)
+ 
+If the PR title is prefixed with any of the default types (`feat`, `fix`, `docs`, `style`, `refactor`, `test`, `chore`) or any of the user configured types as described below (non-case sensitive) then the plugin marks its check as **Success**, or as **Failure** otherwise.
```
MatousJobanek: Here, I would probably also attach a mention that we were inspired by the https://seesparkbox.com/foundry/semantic_commit_messages page - there is also the explanation of all types

docs/fragments/pr-sanitizer.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+To configure PR Sanitizer plugin place `pr-sanitizer.yml` (or `pr-sanitizer.yaml`) file inside of the directory `.ike-prow/` in your project and use properties described below.
+
+.pr-sanitizer.yaml
+[source, yml, indent=0]
+----
+include::../../pkg/plugin/pr-sanitizer/test_fixtures/github_calls/pr-sanitizer.yml[]
+----
+
+<1> Defines set of custom types which will be used to match titles prefixes in changeset and determine if Pull Requests Title conforms with the semantic message style. (Eg. above code snippet configures emoji style semantic commit message.)
```
MatousJobanek: make "emoji style" a link to https://medium.com/walmartlabs/semantic-commit-messages-with-emojis-dba2541cea9a so the user knows what we are talking about :-)

pkg/plugin/pr-sanitizer/event_handler.go
```diff
@@ -0,0 +1,109 @@
...
+
+	default:
+		log.Warnf("received an event of type %q but didn't ask for it", eventType)
+	}
+	return nil
+}
+
+// HasTitleWithValidType checks if title prefix conforms with semantic message style
+func (gh *GitHubLabelsEventsHandler) HasTitleWithValidType(log log.Logger, change scm.RepositoryChange, title string) bool {
+	title = gh.trimWorkInProgressPrefix(log, change, title)
```
MatousJobanek: I guess that in most of the time the PR won't contain the WIP prefix, so I would make it lazy to not download the wip config file every time the sanitizer is invoked. In other words, download the config file and try to trim the prefix only when the first attempt of the title verification fails.

pkg/plugin/pr-sanitizer/event_handler.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+	}
+	return false
+}
+
+func (gh *GitHubLabelsEventsHandler) trimWorkInProgressPrefix(log log.Logger, change scm.RepositoryChange, title string) string {
+	wipPluginConfig := wip.LoadConfiguration(log, change)
+	wipHandler := &wip.GitHubWIPPRHandler{Client: gh.Client, BotName: gh.BotName}
+
+	if ok, prefix := wipHandler.HasWorkInProgressPrefix(title, wipPluginConfig); ok {
```
MatousJobanek: Looking at the implementation of `HasWorkInProgressPrefix` method - the method doesn't use anything from the `GitHubWIPPRHandler` instance, so I think that we can make it a function and call it without instantiating the handler

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -163,7 +163,7 @@ func (gh *GitHubWIPPRHandler) hasWorkInProgressLabel(labels []*gogh.Label, wipLa
 }
 
 // HasWorkInProgressPrefix checks if title is marked as Work In Progress
-func (gh *GitHubWIPPRHandler) HasWorkInProgressPrefix(title string, config PluginConfiguration) (bool, string) {
+func HasWorkInProgressPrefix(title string, config PluginConfiguration) (bool, string) {
```
dipak-pawar: I would make this func definition as `func HasWorkInProgressPrefix(title string, config PluginConfiguration) (string, bool)`, so even if you can use `prefix, ok := wip.HasWorkInProgressPrefix(*event.PullRequest.Title, wip.LoadConfiguration(log, change))`. current use of this function looks little weired. - https://github.com/arquillian/ike-prow-plugins/pull/169/files#diff-4228ddbb4e17d1865d8e954407a5fa84R71

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -172,10 +172,10 @@ func (gh *GitHubWIPPRHandler) HasWorkInProgressPrefix(title string, config Plugi
 			prefixes = config.Prefix
 		}
 	}
-	return gh.hasPrefix(title, prefixes)
+	return hasPrefix(title, prefixes)
 }
 
-func (gh *GitHubWIPPRHandler) hasPrefix(title string, prefixes []string) (bool, string) {
+func hasPrefix(title string, prefixes []string) (bool, string) {
```
dipak-pawar: Same like above

pkg/plugin/pr-sanitizer/event_handler.go
```diff
@@ -0,0 +1,102 @@
...
+	if len(config.TypePrefix) != 0 {
+		if config.Combine {
+			prefixes = append(prefixes, config.TypePrefix...)
+		} else {
+			prefixes = config.TypePrefix
+		}
+	}
+
+	for _, prefix := range prefixes {
+		if strings.HasPrefix(strings.ToLower(title), strings.ToLower(prefix)) {
```
dipak-pawar: Missing TrimSpace for Title which will fail this condition for `title : = " :test: configures plugin"`

so this should be`strings.HasPrefix(strings.ToLower(strings.TrimSpace(title)), strings.ToLower(strings.TrimSpace(prefix)))` I would trim both befre checking `hasPrefix`

pkg/plugin/work-in-progress/event_handler.go
dipak-pawar: The way how are we using as referred in above comment - https://github.com/arquillian/ike-prow-plugins/pull/169/files#diff-4228ddbb4e17d1865d8e954407a5fa84R71, like
`ok, prefix := wip.HasWorkInProgressPrefix(*event.PullRequest.Title, wip.LoadConfiguration(log, change));` is somehow different than go way. Because I would expect `ok` to be as second parameter. I am fine with it if we rename `ok` to some meaningful variable like `prefixed`

MatousJobanek: >  is somehow different than go way. Because I would expect ok to be as second parameter.

this would be valid if the method had a name `GetWorkInProgressPrefix` or `SeparateWorkInProgressPrefix`. However, as we were focused on checking the existence of the prefix from the very beginning and getting the separated prefix is just a side effect, then having it in the current way makes sense. 
So, to solve this situation - how about renaming it to `GetWorkInProgressPrefix` with the return values `(string, bool)`...? Does it make sense to both of you @dipak-pawar @hemanik ?

dipak-pawar: @MatousJobanek +1 for your suggestion

MatousJobanek:
[test]

PR #171, opened 2018-05-31
refactor(#122): make plugin hint effective for all plugins 

### LINKED ISSUE #6: PR title verification (semantic messages)

### LINKED ISSUE #122: Make plugin_hint effective for all plugins
At this moment we only support plugin hints for Test Keeper plugin and the logic how we construct messages is living in the plugin package itself.

As we introduced `plugin_hint` concept to the global plugin configuration we have to enhance this functionality to be available for other plugins too. 

- move the comment logic which can be re-used to its own package
- keep test-keeper specific messages where they are now and use improved comment service to add them to the PR
- consider default messages for each plugin to leverage [golang `template`] (https://golang.org/pkg/text/template/) and have files stored like other assets - if needed, as not all the plugins will need any hints.
- stick to the convention of `.ike-prow/pluginname_hint.md` as the custom hint location. other values should be treated like inline comments

#### Concrete usage

Add a generic comment to WIP plugin explaining what it does and how to use it.

hemanik:
Extracts create comment logic to its own package and create the default message for the plugin using Golang templates.


MatousJobanek:
Hi @hemanik thanks for this PR. To be honest, I don't think that it is necessary to provide a hint message for the `work-in-progress` plugin. it is quite obvious - there is also a label added to the PR, so no additional message is necessary. 
The main task was to make the logic generic so it is easy to add a hint message for example in the `pr-sanitizer` plugin

hemanik:
@MatousJobanek, add the hint for `work-in-progress` since the issue description stated.

> Concrete usage
Add a generic comment to WIP plugin explaining what it does and how to use it.

Can remove it if it's not required.

MatousJobanek:
Yeah, the description in the issue is not clear, I had a discussion with Bartosz about it but forgot to add it to the issue - my bad, sorry.
You can remove it as for the `work-in-progress` plugin it doesn't really bring any additional value.

MatousJobanek:
The main target was that if any new plugin (the `pr-sanitizer`) needs the hint message then the usage should be as easy as possible

alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/arquillian/ike-prow-plugins/6161679d1b512c76884ddf1f140ea447e4cb5da4/.ike-prow/test-keeper.yaml).

pkg/assets/config/message-with-file-not-found.txt
```diff
@@ -0,0 +1,7 @@
+{{.Thumbnail}}
```
MatousJobanek: could you please name this file as `message-with-hint-file-not-found.txt` to make it clear that it is related to the hint file

pkg/comment/comment_message_loader.go
```diff
@@ -0,0 +1,104 @@
...
+		msg = loadMessageTemplate("message-with-no-config.txt")
+	} else {
+		msg = loadMessageTemplate("message-with-config.txt")
+	}
+	return getMsgFromTemplate(msg)
+}
+
+func getMsgFromTemplate(msg string) string {
+	var tpl bytes.Buffer
+	msgTemplate := template.Must(template.New("message").Parse(msg))
```
MatousJobanek: Please, avoid using `*must*` methods that fail with a panic - then, we have no way to easily get the reason of the failure - it is not sent to the Sentry. In addition `template.Must` does nothing else than just checking the error and failing with panic if error is not null:
```go
func Must(t *Template, err error) *Template {
	if err != nil {
		panic(err)
	}
	return t
}
```

pkg/comment/comment_message_loader.go
```diff
@@ -0,0 +1,104 @@
...
+// Message struct for the plugin comment service
+type Message struct {
+	Thumbnail      string
+	Description    string
+	ConfigFile     string
+	Documentation  string
+	MessageFileURL string
+}
+
+var commentMessage Message
```
MatousJobanek: I guess, that this should be used either as a function parameter or the functions as a methods with this receiver. In otrer words, don't use it as a global variable

MatousJobanek: Looking at the line:
https://github.com/arquillian/ike-prow-plugins/pull/171/files#diff-1d8831f0937e79f278010cd60a0e253cR31
the method with a receiver would be probably better

pkg/comment/comment_message_loader_test.go
```diff
@@ -0,0 +1,348 @@
...
+			conf := config.PluginConfiguration{PluginName: ProwPluginName}
+			comment := &comment.Comment{
+				Message: &comment.Message{ConfigFile: conf.LocationURL, Documentation: "#_my_test_plugin"},
+			}
+
+			// when
+			msg := comment.LoadMessage(conf, scm.RepositoryChange{})
+
+			// then
+			Expect(msg).To(ContainSubstring("http://arquillian.org/ike-prow-plugins/#_my_test_plugin"))
```
MatousJobanek: verify that it doesn't contain the string `Your plugin configuration is stored in the` - opposite to this https://github.com/arquillian/ike-prow-plugins/pull/171/files#diff-5a1174093f81d3fdc9e33c0c3727465dR46

pkg/comment/comment_message_loader_test.go
```diff
@@ -0,0 +1,348 @@
...
+			Expect(msg).To(ContainSubstring(url))
+		})
+	})
+
+	Context("Creation of default comment messages from default location when config plugin hint is not set", func() {
+
+		BeforeEach(func() {
+			defer gock.OffAll()
+		})
+
```
MatousJobanek: Don't forget to call `AfterEach(EnsureGockRequestsHaveBeenMatched)`

pkg/comment/comment_message_loader.go
```diff
@@ -0,0 +1,111 @@
...
+	assets "github.com/arquillian/ike-prow-plugins/pkg/assets/generated"
+	"github.com/arquillian/ike-prow-plugins/pkg/config"
+	"github.com/arquillian/ike-prow-plugins/pkg/github/service"
+	"github.com/arquillian/ike-prow-plugins/pkg/log"
+	"github.com/arquillian/ike-prow-plugins/pkg/scm"
+	"github.com/arquillian/ike-prow-plugins/pkg/utils"
+)
+
+// Comment struct for the plugin comment service
+type Comment struct {
```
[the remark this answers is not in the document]
MatousJobanek: and do the same also with the files in the `test-keeper` plugin. all the logic is related to the hints if I'm not wrong...

hemanik:
Kept the `comment_message_test.go` in the `test-keeper plugin` to verify the plugin-specific messages. Will remove it if you feel it is not required.

pkg/plugin/test-keeper/hint_message.go
```diff
@@ -0,0 +1,36 @@
...
+	hint := &hint.Hint{
+		Log: log,
+		Message: &hint.Message{
+			Thumbnail:     sadIke,
+			Description:   beginning,
+			ConfigFile:    configuration.LocationURL,
+			Documentation: documentationSection,
+		},
+	}
+	return hint.LoadMessage(configuration.PluginConfiguration, change)
```
dipak-pawar: I would avoid variable name to be the same as a package which we are importing https://github.com/arquillian/ike-prow-plugins/pull/171/files#diff-78aa0452e886a8e2e8f9d279288827fbR4

PR #179, opened 2018-06-07
chore(#181): listening on 8888 for /metrics in prow+pmcd.sh

dipak-pawar:
**Changes Proposed**
 Change `'http://localhost:8080/metrics' > $URLSD/prow.url` to `'http://localhost:8888/metrics' > $URLSD/wit.url`

Fixes: #181

cc @fche


MatousJobanek:
why `wit.url`? shouldn't that be `prow.url`?

dipak-pawar:
@MatousJobanek I have added which is suggested by @fche. Maybe @fche has answer for it. But my guess is they are using the same url(`wit.url`) in collector service which is running across cluster. Also  you can see same url for other services([auth](https://github.com/fabric8-services/fabric8-auth/blob/master/auth%2Bpmcd.sh#L38)).

MatousJobanek:
Ok, thanks for explanation @dipak-pawar

MatousJobanek:
[test]

fche:
The 'prow' name can stay too; it's just that the default html dashboard that aggregates osio components happens to use wildcards with *.prometheus.wit.* in them.  We could use 'prow' for these, and then also have to update the dashboard farther.  No big deal either way.

MatousJobanek:
[test]

fche:
lgtm

dipak-pawar:
[test]

MatousJobanek:
[test]

Issue #186, opened 2018-06-11
Add TestKeeper Plugin Metrics


dipak-pawar:
Add following custom metrics for Test keeper
- /ok-without-tests vs size of the PR
- PRs with and without tests

PR #187, opened 2018-06-11
feat(#186): implement custom metrics for test-keeper

dipak-pawar:
Fixes: #186


pkg/plugin/plugin_init.go
```diff
@@ -95,15 +99,13 @@ func InitPlugin(pluginName string, newEventHandler EventHandlerCreator, newServe
...
-		}
-		errLog.Fatal("Prometheus metrics registration failed!")
-	}
+	logErrors(errs, logger, "Prometheus metrics registration failed!")
+
 	pluginServer.Metrics = metrics
 
+	errors := registerMetrics()
+	logErrors(errors, logger, fmt.Sprintf("Prometheus metrics registration failed for plugin %s!", pluginName))
+
```
MatousJobanek: is it possible to register the plugin specific metrics as part of the service creation?
there would be a few benefits
* we would get rid of calling similar method/function twice
* we would need to have the additional method parameter
* the plugins that don't use any metric wouldn't need to create the dummy function: https://github.com/arquillian/ike-prow-plugins/pull/187/files#diff-3bfa5d1e5db805dd35fac5a9d5724c61R51

pkg/plugin/test-keeper/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+		Help:    "Histogram for changed file size of pull request with ok-without-tests.",
+		Buckets: prometheus.ExponentialBuckets(1, 5, 6),
+	}, []string{"full_name"})
+)
+
+// RegisterMetrics registers prometheus collectors to collect metrics for test-keeper.
+func RegisterMetrics() []error {
+	errors := make([]error, 0, 2)
+	server.RegisterOrGet(pullRequestsCounter, &errors, func(collector prometheus.Collector) {
+		pullRequestsCounter = collector.(*prometheus.CounterVec)
```
MatousJobanek: I'm a bit confused now. In the general `metrics` https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/server/metrics.go#L32 we assign it to the variable inside of the object. Here it is assigned to a global variable. 
I'd like to have it consistent or is there any difference in the usage?

pkg/plugin/test-keeper/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+func reportPullRequest(l log.Logger, pr *gogh.PullRequest, prType string) {
+	fullName := *pr.Head.Repo.FullName
+	if counter, err := pullRequestsCounter.GetMetricWithLabelValues(fullName, prType); err != nil {
+		l.Errorf("Failed to get pull request metric for Repository: %q. Cause: %q", fullName, err)
+	} else {
+		counter.Inc()
+	}
+}
+
+func reportOkWithoutTestsPullRequest(pr *gogh.PullRequest) {
```
MatousJobanek: I would call it `reportOkWithoutTestsComment` or `reportBypassCommand` as this expresses what the method really represents - reporting when the bypass command is used

pkg/plugin/test-keeper/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+	WithTests = "with_tests"
+	// WithoutTests is a label used in prometheus metrics pull_requests_total for pull request without any tests.
+	WithoutTests = "without_tests"
+)
+
+var (
+	pullRequestsCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "pull_requests_total",
+		Help: "Total number of pull requests.",
+	}, []string{"full_name", "type"})
```
MatousJobanek: I'd like to welcome some information about the plugin if it is plugin-specific metric

pkg/plugin/test-keeper/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+	okWithoutTestsPullRequest = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "pull_requests_changed_file_size",
```
MatousJobanek: `pull_requests_changed_file_count` or `pull_requests_size` 
WDYT?

pkg/plugin/test-keeper/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+
+	server.RegisterOrGet(okWithoutTestsPullRequest, &errors, func(collector prometheus.Collector) {
+		okWithoutTestsPullRequest = collector.(*prometheus.HistogramVec)
+	})
+
+	return errors
+}
+
+func reportPullRequest(l log.Logger, pr *gogh.PullRequest, prType string) {
+	fullName := *pr.Head.Repo.FullName
```
MatousJobanek: This is the label that you use for collecting the counter? Correct me if I'm wrong - this is the repository name the PR was sent from. Could you please explain to me what you are trying to achieve here? From the code, I have a feeling that you count every PR event that is received in `test-keeper` plugin for repository the PR has been sent from separately.

MatousJobanek: similarly here https://github.com/arquillian/ike-prow-plugins/pull/187/files#diff-241f68d9d4673ceec43fbd014ca925acR53

pkg/plugin/test-keeper/metrics_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+		gock.New("https://api.github.com").
+			Post("/repos/" + repositoryName + "/statuses").
+			SetMatcher(ExpectPayload(toHaveEnforcedSuccessState)).
+			Reply(201) // This way we implicitly verify that call happened after `HandleEvent` call
+
+		statusPayload := LoadFromFile("test_fixtures/github_calls/prs/without_tests/skip_comment_by_admin.json")
+
+		// when
+		err := handler.HandleEvent(log, github.IssueComment, statusPayload)
```
MatousJobanek: try to invoke the handler several times with different type of the payload. This would give me more information what is expected from the logic

pkg/plugin/plugin_init.go
dipak-pawar: I didn't get your point, what do you mean by service creation?

pkg/plugin/test-keeper/metrics.go
dipak-pawar: I was thinking about the same, but it's not size in `bytes` or some units. so it's quite difficult to understand while reading it.

pkg/plugin/test-keeper/metrics.go
dipak-pawar: I think you can select host in PCP and graphana charts due to which I believe this is not needed

pkg/plugin/test-keeper/metrics.go
dipak-pawar: this is not a global variable, we are assigning it as an unexported variable. no there is nothing difference in usage.

pkg/plugin/plugin_init.go
MatousJobanek: In the implementation of the `ServerCreator` https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/plugin/plugin_init.go#L53 that is called here https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/plugin/plugin_init.go#L96

pkg/plugin/test-keeper/metrics.go
MatousJobanek: unexported global variable :-). global variable in the context of this file - never mind - the main point was that the code/usage is inconsistent

pkg/plugin/test-keeper/metrics.go
MatousJobanek: what if I don't have it filtered (when I display all graphs) - then it is visible as well, right? Then it would make sense adding the plugin name to the metric name and its description

pkg/plugin/test-keeper/metrics.go
MatousJobanek: I think that it is clear in the context of PR that size refers to the number of changed files. Personally, I would vote for `pull_requests_size`

pkg/server/metrics.go
```diff
@@ -19,31 +19,32 @@ var (
...
 
-	return metrics, errors
+	return errors
 }
 
-func registerOrGet(collector prometheus.Collector, errors *[]error, assign func(regCollector prometheus.Collector)) {
+// RegisterOrGet registers the provided Collector with the DefaultRegisterer and
+// assigns the Collector, unless an equal Collector was registered before, in
+// which case that Collector is assigned.
+func RegisterOrGet(collector prometheus.Collector, errors *[]error, assign func(regCollector prometheus.Collector)) {
```
MatousJobanek: just please remind me why the method is named `RegisterOrGet` and not `RegisterMetric` - I'm not sure if we have talked about it or not...

pkg/plugin/test-keeper/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+	WithTests = "with_tests"
+	// WithoutTests is a label used in prometheus metrics pull_requests_total for pull request without any tests.
+	WithoutTests = "without_tests"
+)
+
+var (
+	pullRequestsCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "pull_requests_total",
+		Help: "Total number of pull requests.",
+	}, []string{"full_name", "type", "plugin_name"})
```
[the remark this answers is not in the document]
dipak-pawar: I am not sure why it's needed to have plugin name to be added. If you can see in following screenshot you can easily see for which plugin this particular metrics belongs to

screenshot from 2018-06-19 11-10-39

This screenshot captured for all hosts. When you select for single host, you can see that particulat metrics. 

If particular metric (`pull_requests_total`) is only for single host, then graphana will show only for that host which is the case here for `pull_requests_total` and `okWithoutTestsPullRequest`

pkg/server/metrics.go
dipak-pawar: I think I forgot to rename it after lots of refactoring for first metrics pr. Renaming to `RegisterOrAssignCollector`

pkg/plugin/test-keeper/metrics.go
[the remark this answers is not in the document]
MatousJobanek: so, you are saying that Graphana doesn't show these test-keeper metrics when there is no filtering set - in other words, when obrazek ?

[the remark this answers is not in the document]
dipak-pawar: I didn't mean that. When you have `host: All`, and you are looking for test keeper metrics e.g. `pull_requests_total` (which is only for `test-keeper` not for `wip` and `pr-sanitizer`), so in this case only test-keeper `pull_requests_total` metrics will show.

If we have `host:All`, and we are looking for `total_handled_events` which is there for all plugins it will show all three in graph, and if you hover over it you can easily see for which plugin it belongs to

Issue #27, opened 2018-03-06
Change "WIP" status based on existing label


bartoszmajsak:

hemanik:
> Question: If we remove the label should we remove "WIP" suffix to (and the other way around)?

The proposed solution assumes 4 different cases described below that are handled for WIP PR.

*Case#1:* The PR is has a `WIP` label but the title lacks `WIP` prefix.
- The PR status is failed and the title is left as it is.

*Case#2:* The PR title contains the `WIP` prefix but no label.
- The status is failed and the label is added.

*Case#3:* The PR title is updated and `WIP` prefix is removed.
- The status is changed to success and the label is removed.

*Case#4:* The PR is updated and the `WIP` label is removed.
- The status is changed to success and the prefix is removed.

Issue #57, opened 2018-03-22
Ready-to-review label


bartoszmajsak:
When all statuses are green PR should be marked as ready-to-peer-review so that other developer can start reviewing the code without being distracted by issues which automated tools can find.

PR #189, opened 2018-06-12
 feat(#27): change status based on existing label

hemanik:
The current WIP plugin looks for the PR title prefix to set the status. The scope of this enhanced to change PR status based on the existing label.

The new behavior handles the following cases as described below:

- set the status to failure if PR either contains a `wip` label (irrespective to the title prefix) or it has a `wip` prefix in the title (in which case the label is automatically added).
- set the status to success if the PR is updated by either removing the `wip` prefix from the title (in which case the label is also removed automatically) or by removing the `wip` label (in which case, if present, the `wip` title prefix is also remove automatically).

Fixes: #27


pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -178,3 +146,89 @@ func (gh *GitHubWIPPRHandler) checkTitleAndSetStatus(log log.Logger, pullRequest
...
+	for _, prefix := range prefixes {
+		pattern := `(?mi)^(\[|\()?` + prefix + `(\]|\))?(:| )+`
+		if match, _ := regexp.MatchString(pattern, title); match {
+			return prefix, true
+		}
+	}
+	return "", false
+}
+
+func (gh *GitHubWIPPRHandler) getRepositoryChange(pullRequest *gogh.PullRequest) scm.RepositoryChange {
```
MatousJobanek: This should be placed next to the `scm.RepositoryChange` struct as its constructor. And by convention, the name should start with "New"

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -178,3 +146,89 @@ func (gh *GitHubWIPPRHandler) checkTitleAndSetStatus(log log.Logger, pullRequest
...
+	statusService := gh.getStatusService(log, change)
+	labels := gh.getPullRequestLabels(log, change, pullRequest)
+	configuration := LoadConfiguration(log, change)
+
+	labelExists := gh.hasWorkInProgressLabel(labels, configuration.Label)
+	if labelExists {
+		return statusService.Failure(InProgressMessage, InProgressDetailsPageName)
+	}
+	return statusService.Success(ReadyForReviewMessage, ReadyForReviewDetailsPageName)
+}
```
MatousJobanek: the content of the methods `updateTitleAndResetStatus` and `checkLabelAndSetStatus` seem to be pretty similar. could we merge them?

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -178,3 +146,89 @@ func (gh *GitHubWIPPRHandler) checkTitleAndSetStatus(log log.Logger, pullRequest
 	return statusService.Success(ReadyForReviewMessage, ReadyForReviewDetailsPageName)
 
 }
+
+func (gh *GitHubWIPPRHandler) updateTitleAndResetStatus(log log.Logger, pullRequest *gogh.PullRequest) error {
+	change := gh.getRepositoryChange(pullRequest)
+	statusService := gh.getStatusService(log, change)
+	labels := gh.getPullRequestLabels(log, change, pullRequest)
```
MatousJobanek: Looking at the  `PullRequestEvent` struct
```go
type PullRequestEvent struct {
	// Action is the action that was performed. Possible values are:
	// "assigned", "unassigned", "review_requested", "review_request_removed", "labeled", "unlabeled",
	// "opened", "closed", "reopened", "synchronize", "edited".
	// If the action is "closed" and the merged key is false,
	// the pull request was closed with unmerged commits. If the action is "closed"
	// and the merged key is true, the pull request was merged.
	Action      *string      `json:"action,omitempty"`
	Number      *int         `json:"number,omitempty"`
	PullRequest *PullRequest `json:"pull_request,omitempty"`

	// The following fields are only populated by Webhook events.
	Changes *EditChange `json:"changes,omitempty"`
	// RequestedReviewer is populated in "review_requested", "review_request_removed" event deliveries.
	// A request affecting multiple reviewers at once is split into multiple
	// such event deliveries, each with a single, different RequestedReviewer.
	RequestedReviewer *User         `json:"requested_reviewer,omitempty"`
	Repo              *Repository   `json:"repository,omitempty"`
	Sender            *User         `json:"sender,omitempty"`
	Installation      *Installation `json:"installation,omitempty"`
	Label             *Label        `json:"label,omitempty"` // Populated in "labeled" event deliveries.
}
```
at the very end, there is the Label variable? could we use it?

MatousJobanek: And if not from this place - there is a variable `Labels  []*Label` in the `PullRequest` struct, so there shouldn't be no need to retrieve it by sending another request.

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -178,3 +146,89 @@ func (gh *GitHubWIPPRHandler) checkTitleAndSetStatus(log log.Logger, pullRequest
...
+	configuration := LoadConfiguration(log, change)
+
+	labelExists := gh.hasWorkInProgressLabel(labels, configuration.Label)
+	if !labelExists {
+		if prefix, ok := gh.IsWorkInProgress(*pullRequest.Title, configuration); ok {
+			*pullRequest.Title = strings.TrimPrefix(*pullRequest.Title, prefix)
+			err := gh.Client.EditPullRequest(change, *pullRequest.Number, pullRequest)
+			if err != nil {
+				log.Warnf("failed to update PR title [%q]. cause: %s", *pullRequest, err)
```
MatousJobanek: that should be logged as an error

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -178,3 +146,89 @@ func (gh *GitHubWIPPRHandler) checkTitleAndSetStatus(log log.Logger, pullRequest
...
+			}
+		}
+		return statusService.Success(ReadyForReviewMessage, ReadyForReviewDetailsPageName)
```
MatousJobanek: if the label was removed, but the edit of the PR [failed](https://github.com/arquillian/ike-prow-plugins/pull/189/files#diff-11865c15f9f479c8ca04795d06b465acR161) should we change the status to success as well? then it looks a bit inconsistent - the PR contains the prefix, but the status is green...
maybe, instead of logging the error https://github.com/arquillian/ike-prow-plugins/pull/189/files#diff-11865c15f9f479c8ca04795d06b465acR162 we should return it

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -178,3 +146,89 @@ func (gh *GitHubWIPPRHandler) checkTitleAndSetStatus(log log.Logger, pullRequest
...
+	if labelExists {
+		return statusService.Failure(InProgressMessage, InProgressDetailsPageName)
+	}
+	return statusService.Success(ReadyForReviewMessage, ReadyForReviewDetailsPageName)
+}
+
+func (gh *GitHubWIPPRHandler) getPullRequestLabels(log log.Logger, change scm.RepositoryChange, pullRequest *gogh.PullRequest) []*gogh.Label {
+	labels, err := gh.Client.ListPullRequestLabels(change, *pullRequest.Number)
+	if err != nil {
+		log.Warnf("failed to list labels on PR [%q]. cause: %s", *pullRequest, err)
```
MatousJobanek: log it as an error.
btw, I think that you should call `err.Error()` to get the message

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -178,3 +146,89 @@ func (gh *GitHubWIPPRHandler) checkTitleAndSetStatus(log log.Logger, pullRequest
...
+	for _, label := range labels {
+		if label.GetName() == wipLabel {
+			return true
+		}
+	}
+	return false
+}
+
+// IsWorkInProgress checks if title is marked as Work In Progress
+func (gh *GitHubWIPPRHandler) IsWorkInProgress(title string, config PluginConfiguration) (string, bool) {
```
MatousJobanek: with the ability to use labels, we should rename the method for example to: `HasWorkInProgressPrefix`

MatousJobanek: And I would probably swap the return parameters - the first would be bool, then the prefix. The reason is that the main purpose of this method is checking if the title contains the prefix, the retrieved prefix is just a side effect.

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -178,3 +146,89 @@ func (gh *GitHubWIPPRHandler) checkTitleAndSetStatus(log log.Logger, pullRequest
...
+		}
+	}
+	return gh.hasPrefix(strings.ToLower(title), prefixes)
+}
+
+func (gh *GitHubWIPPRHandler) hasPrefix(title string, prefixes []string) (string, bool) {
+	for _, prefix := range prefixes {
+		pattern := `(?mi)^(\[|\()?` + prefix + `(\]|\))?(:| )+`
+		if match, _ := regexp.MatchString(pattern, title); match {
+			return prefix, true
```
MatousJobanek: what if the prefix contains also brackets? are they matched and then removed as well? the same applies also to the colon.

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -178,3 +146,89 @@ func (gh *GitHubWIPPRHandler) checkTitleAndSetStatus(log log.Logger, pullRequest
...
+
+func (gh *GitHubWIPPRHandler) getRepositoryChange(pullRequest *gogh.PullRequest) scm.RepositoryChange {
+	return scm.RepositoryChange{
+		Owner:    *pullRequest.Base.Repo.Owner.Login,
+		RepoName: *pullRequest.Base.Repo.Name,
+		Hash:     *pullRequest.Head.SHA,
+	}
+}
+
+func (gh *GitHubWIPPRHandler) getStatusService(log log.Logger, change scm.RepositoryChange) scm.StatusService {
```
MatousJobanek: name it `newStatusService`

pkg/plugin/work-in-progress/event_handler.go
MatousJobanek: there is already such a function: https://github.com/arquillian/ike-prow-plugins/blob/master/pkg/github/service/repository_change.go so please use this one

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -146,35 +119,69 @@ func (gh *GitHubWIPPRHandler) handlePrComment(log log.Logger, comment *gogh.Issu
...
-			if err := gh.Client.AddPullRequestLabel(change, *pullRequest.Number, strings.Fields(configuration.Label)); err != nil {
-				log.Errorf("failed to add label on PR [%q]. cause: %s", *pullRequest, err)
+	labelExists := gh.hasWorkInProgressLabel(pullRequest.Labels, configuration.Label)
+	prefixExists, prefix := gh.HasWorkInProgressPrefix(*pullRequest.Title, configuration)
+
+	if prefixExists && !labelExists {
+		if labelUpdated {
+			*pullRequest.Title = strings.TrimPrefix(*pullRequest.Title, prefix)
+			if err := gh.Client.EditPullRequest(pullRequest); err != nil {
+				return fmt.Errorf("failed to update PR title [%q]. cause: %s", *pullRequest, err)
```
MatousJobanek: Correct me if I'm wrong - to get the nice format of the error (that contains a message of the original error) it is necessary to call `err.Error()` on the original error - that retrieves the message of the root cause.

pkg/plugin/work-in-progress/event_handler.go
```diff
@@ -146,35 +119,69 @@ func (gh *GitHubWIPPRHandler) handlePrComment(log log.Logger, comment *gogh.Issu
...
 			}
+			return statusService.Success(ReadyForReviewMessage, ReadyForReviewDetailsPageName)
+		}
+		if err := gh.Client.AddPullRequestLabel(change, *pullRequest.Number, strings.Fields(configuration.Label)); err != nil {
```
MatousJobanek: why `strings.Fields`? the label should be only one and can also contain a white space - eg: `work in progress`

pkg/plugin/work-in-progress/event_handler_unit_test.go
```diff
@@ -20,7 +20,8 @@ var _ = Describe("Work-in-progress Plugin features", func() {
 
 		DescribeTable("should recognize PR as work-in-progress if title starts with configured or default prefix",
 			func(title string) {
-				Expect(handler.IsWorkInProgress(title, wip.PluginConfiguration{Prefix: []string{`On Hold`}, Combine: true})).To(BeTrue())
+				state, _ := handler.HasWorkInProgressPrefix(title, wip.PluginConfiguration{Prefix: []string{`On Hold`}, Combine: true})
```
MatousJobanek: could you please also verify (in all cases) the retrieved prefix - this way you verify that the [find](https://github.com/arquillian/ike-prow-plugins/pull/189/files#diff-11865c15f9f479c8ca04795d06b465acR182) works correctly

pkg/github/client/client.go
```diff
@@ -28,6 +29,7 @@ type Client interface {
 	ListPullRequestLabels(change scm.RepositoryChange, prNumber int) ([]*gogh.Label, error)
```
MatousJobanek: is this used anywhere?

pkg/plugin/work-in-progress/event_handler_gh_test.go
```diff
@@ -49,6 +50,68 @@ var _ = Describe("Work In Progress Plugin features", func() {
...
+		It("should mark opened PR as work-in-progress when labeled with WIP", func() {
+			// given
+			NonExistingRawGitHubFiles("work-in-progress.yml", "work-in-progress.yaml")
+
+			gock.New("https://api.github.com").
+				Get("/repos/bartoszmajsak/wfswarm-booster-pipeline-test/issues/4/labels").
+				Reply(200).
+				BodyString(`[{"id": 934813958,` +
+					`"url": "https://api.github.com/repos/bartoszmajsak/wfswarm-booster-pipeline-test/labels/work-in-progress",` +
+					`"name": "work-in-progress", "color": "ededed", "default": false}]`)
```
MatousJobanek: this mock (and all other mocks for the labels listing) is not necessary - the information is taken from the PR

pkg/plugin/work-in-progress/event_handler_gh_test.go
```diff
@@ -49,6 +50,68 @@ var _ = Describe("Work In Progress Plugin features", func() {
...
+			NonExistingRawGitHubFiles("work-in-progress.yml", "work-in-progress.yaml")
+
+			gock.New("https://api.github.com").
+				Get("/repos/bartoszmajsak/wfswarm-booster-pipeline-test/issues/4/labels").
+				Reply(200).
+				BodyString("[]")
+
+			gock.New("https://api.github.com").
+				Patch("repos/bartoszmajsak/wfswarm-booster-pipeline-test/pulls/4").
+				Reply(200).
```
MatousJobanek: please add also a matcher checking that the title was correctly updated - removed the prefix including the flavours

pkg/plugin/work-in-progress/event_handler.go
hemanik: The following function used to add label expects a string array of labels. We only have one string label and that needs to be converted to a string array and hence the choice of `string.Field()`. 

The choice of converting the string to string array is done at this location instead inside the  `gh.Client.AddPullRequestLabel` because I wanted to keep this method unmodified considering this could be used elsewhere where we could have to send multiple label values. (Example: https://github.com/arquillian/ike-prow-plugins/issues/57) 
  
```
// AddLabelsToIssue adds labels to an issue.
//
// GitHub API docs: https://developer.github.com/v3/issues/labels/#add-labels-to-an-issue
func (s *IssuesService) AddLabelsToIssue(ctx context.Context, owner string, repo string, number int, labels []string) ([]*Label, *Response, error) {
```

hemanik: Thanks for the explanation @MatousJobanek, updated it.

docs/fragments/status/work-in-progress/success/wip-success.adoc
```diff
@@ -1,6 +1,6 @@
 ==== Success [[wip-success]]
 
-Your Pull Request has been approved as the plugin didn't detect any sign of work in progress (the PR title doesn't start with any of the <<index#work-in-progress-config,"work in progress prefixes">>).
+Your Pull Request has been approved as the plugin didn't detect any sign of work in progress (the PR title doesn't start with any of the <<index#work-in-progress-config,"work in progress prefixes">> or contain any "work-in-progress" label).
```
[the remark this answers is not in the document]
MatousJobanek: I see that you have updated the failure page. https://github.com/arquillian/ike-prow-plugins/pull/189/commits/39dc7fab0b670d05de519720f934c03afb5d1391
I wanted to say that you should update the success page to be same as the failure page was, not the other way around. I see that my previous comment is very confusing and also grammatically wrong, so I do not wonder that you misunderstood. Sorry for that :-) 
Could you please change both of them one more time? Thanks :-)

MatousJobanek:
[test]

PR #190, opened 2018-06-12
chore: use pr loader constuctor instead of helper method

MatousJobanek:
This is refactoring change - I've noticed it when I was going through the code, that we can get rid of the helper method and use pr loader constructor that takes an already loaded PR


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/arquillian/ike-prow-plugins/blob/91a001ce681574e029147b508c5a209014e8773c/.ike-prow/test-keeper.yaml).

Issue #204, opened 2018-06-20
Make pr-sanitizer running in cluster


dipak-pawar:
We already have `pr-sanitizer` initial implementation in place. Enable it for our own repository to catch early bug if any.

**Steps to perform:**
- [x] saas-errortracking - add definition for the new service - https://github.com/openshiftio/saas-errortracking/pull/9
- [x] modify cico script to build all three plugins - #206
- [x] create an HK issue for the configmap update - https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2062
- [x] change osio-plugins.yaml - #205

MatousJobanek:
I'm a bit confused. Why the HK issue contains all secrets and configmaps? Correct me if I'm wrong, the only thing that will be necessary to update is the `plugins.yaml` file. Secrets as well as the `config.yaml` file are already there...

MatousJobanek:
and another thing is that the HK issue should have been created after the `plugins.yaml` update https://github.com/arquillian/ike-prow-plugins/pull/205 to contain the updated configuration

dipak-pawar:
@MatousJobanek applied new configuration for pr-sanitizer. now it's enabled for `ike-prow-plugins` repo.

MatousJobanek:
Awesome, good job @dipak-pawar :+1:

PR #203, opened 2018-06-20
feat(#182): adds /run command support for pr-sanitizer.

hemanik:
Adds `/run` command support for pr-sanitizer plugin
- if `/run pr-sanitizer` is added as comment to an issue the pr-sanitizer plugin is triggered.
- if `/run all` is added as comment,  all plugins and necessary checks will be triggered. 

Fixes: #182

PR #205, opened 2018-06-20
chore(#204): enable pr-sanitizer plugin for ike-prow-plugins

dipak-pawar:
This is using following configuration.
```yaml
- name: pr-sanitizer
      events:
        - pull_request
        - issue_comment
```
Currently we don't have `issue_comment` event handled. we should wait for #203 to be merged before we make this upstream as #203 is adding support `issue_comment` event.

Fixes: #204


dipak-pawar:
@MatousJobanek: Sure. you are right, it was not wip. Also I didn't have idea about `the merge of this PR doesn't influence the running instance anyhow until the actual configmap is being updated in OpenShift`, so thanks for clarification.

Issue #202, opened 2018-06-20
Lock prometheus dependency


MatousJobanek:
https://github.com/arquillian/ike-prow-plugins/pull/133#discussion_r184799826

PR #207, opened 2018-06-21
chore(#202): lock prometheus dependency

dipak-pawar:
Fixes: #202


MatousJobanek:
I have one question - Is there any reason for locking only `github.com/prometheus/client_golang` library and not the rest of the premetheus libraries?
* `github.com/prometheus/client_model`
* `github.com/prometheus/common`
* `github.com/prometheus/procfs`

Couldn't that lead to incompatibility issues when the other libraries are changed but `client_goland` stays in the fixed version?

Issue #197, opened 2018-06-14
Refactor EventHandler interface


MatousJobanek:
Currently, we have `EventHandler` interface with one method:
```go
HandleEvent(log log.Logger, eventType github.EventType, payload []byte) error
```
it takes the event and in every plugin implementations it checks for the type of the event and unmarshals to the corresponding object. This behavior is same in all plugin implementations.
To minimize the code duplication change the interface to contain two methods:
```go
HandlePullRequestEvent(log log.Logger, event *gogh.PullRequestEvent) error
HandleIssueCommentEvent(log log.Logger, event *gogh.IssueCommentEvent) error
```
and move the common logic to the `server.go`

PR #208, opened 2018-06-21
refactor(#197): event handler interface to abstract common logic.

hemanik:
This PR refactors Handler interface and related logic in plugins by replacing the existing single method by introducing 2 methods to minimize the code duplication:

```
HandlePullRequestEvent(log log.Logger, event *gogh.PullRequestEvent) error
HandleIssueCommentEvent(log log.Logger, event *gogh.IssueCommentEvent) error
```

Fixes: #197

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 89baf0cd29908af47405714762c68467df13cc85df8f24a51b297c51fc6a051c
-updated: 2018-06-20T17:43:57.704285941+05:30
+hash: ec491e3f9011b7709d9ebc6cddbc6a0d8af5d72ac15e77aa01243f65499cf5f0
+updated: 2018-06-22T14:10:56.338676246+05:30
 imports:
 - name: github.com/bazelbuild/buildtools
   version: a6baf7df1e358187d43bb6b9f3debc6da8454922
@@ -89,7 +89,7 @@ imports:
 - name: github.com/pkg/errors
   version: a22138067af1c4942683050411a841ade67fe1eb
 - name: github.com/prometheus/client_golang
-  version: f504d69affe11ec1ccb2e5948127f86878c9fd57
+  version: faf4ec335fe01ae5a6a0eaa34a5a9333bfbd1a30
   subpackages:
   - prometheus
   - prometheus/promhttp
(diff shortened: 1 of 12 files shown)


pkg/internal/test/test_doubles.go
```diff
@@ -32,6 +34,24 @@ func FromFile(filePath string) io.Reader {
 	return file
 }
 
+// NewIssueCommentEvent creates an instance of gogh.PullRequestEvent with the given values
+func NewIssueCommentEvent(payload []byte) *gogh.IssueCommentEvent {
```
MatousJobanek: Could these two methods take a path to the file and load the content of the file in this method or is there some case when it is better to provide the bytes? If it is possible to load the file here, then I would call the method `LoadIssueCommentEvent`

pkg/plugin/pr-sanitizer/event_handler.go
```diff
@@ -43,48 +42,18 @@ var (
...
+func (gh *GitHubLabelsEventsHandler) HandlePullRequestEvent(log log.Logger, event *gogh.PullRequestEvent) error {
 	if !utils.Contains(handledPrActions, *event.Action) {
 		return nil
 	}
 	return gh.checkTitleAndSetStatus(log, event.PullRequest)
 }
 
-func (gh *GitHubLabelsEventsHandler) handlePrComment(log log.Logger, comment *gogh.IssueCommentEvent) error {
+// HandleIssueCommentEvent is an entry point for the plugin logic. This method is invoked by the Server when
+// events are dispatched from the /hook service
```
MatousJobanek: `This method is invoked by the Server when comment event is dispatched from the /hook service`

glide.lock
```diff
@@ -89,7 +89,7 @@ imports:
 - name: github.com/pkg/errors
   version: a22138067af1c4942683050411a841ade67fe1eb
 - name: github.com/prometheus/client_golang
-  version: faf4ec335fe01ae5a6a0eaa34a5a9333bfbd1a30
+  version: f504d69affe11ec1ccb2e5948127f86878c9fd57
```
MatousJobanek: this version should have been locked https://github.com/arquillian/ike-prow-plugins/blob/master/glide.lock#L92

glide.lock
```diff
@@ -1,5 +1,5 @@
-hash: ec491e3f9011b7709d9ebc6cddbc6a0d8af5d72ac15e77aa01243f65499cf5f0
-updated: 2018-06-21T16:58:05.578562323+05:30
+hash: 89baf0cd29908af47405714762c68467df13cc85df8f24a51b297c51fc6a051c
+updated: 2018-06-20T17:43:57.704285941+05:30
```
MatousJobanek: This is an older version

PR #148, opened 2018-05-16
feat(#26): adds work-in-progress label.

hemanik:
This PR introduces the feature to automatically add and remove `work-in-progress` label to a PR depending on the WIP status as inferred from the PR title. 

The label name can be configured using the `work-in-progress.yaml` config file. If it is not configured the default name `work-in-progress` is used.

Fixes: #26

Issue #145, opened 2018-05-07
Update Ike's reaction in the PR if the state changes


gastaldi:
About the plug-in that checks for tests in a PR, it would be nice if Ike's reaction (in the PR comment) were updated when the PR is updated. 

Even though that the PR status changes accordingly, updating the comment with Ike's feedback would be a nice visual indicator for PR reviewers ;)

maxandersen:
isn't this more apropriately done by a new comment + update via status api ?

gastaldi:
@maxandersen I'd prefer to update the existing comment instead of adding new comments as this can confuse someone when reviewing the PR (specially if it has a lot of comments)

maxandersen:
then I would say better to use status api updates rather than comments. ie. like a failed build or linting check.

  Could also make the icon slightly smaller so it does not take up that amount of visual space :)

gastaldi:
It would be nice to display it in the newest Checks tab: https://help.github.com/articles/about-status-checks/#checks

MatousJobanek:
Hi guys, 

I finally got to this issue - sorry for the delay.

> I'd prefer to update the existing comment instead of adding new comments as this can confuse someone when reviewing the PR (especially if it has a lot of comments)

as for the update of the message hint - we have thee options here:
1. remove the comment and add a new one
2. update the current comment
3. remove the comment and don't add any new one
I can imagine that adding a new comment can be quite annoying - especially when there is some ongoing discussion, so I would vote for either the second or the third option.

> then I would say better to use status api updates rather than comments.

we do use it and we add also details link there to a page that explains what is going on, but nobody has noticed it obviously :-). There are two reasons why we add the comment there.
1. when there is a PR without a test, we want to do the maximum to force the committer to add a test to the PR.
2. we can customize the message taking information from the project/configuration file and also by mentioning the committer there

> Could also make the icon slightly smaller so it does not take up that amount of visual space :)

It is still possible to customize the message in every project separately by using a config file, so you can define your own message, that will be shorter and without any images.
But if you agree on the image being too big, then I can definitely make it smaller ;-)

> It would be nice to display it in the newest Checks tab

First of all, we would have to make the Ike plugins a GH app, then we can use the checks tab. There is an issue for it: https://github.com/arquillian/ike-prow-plugins/issues/59

MatousJobanek:
So, my proposal is:

* If there is no test, then the comment will contain a smaller image and will look like:

[...]

PR #209, opened 2018-06-21
feat(#148): rework hint to status message that can be updated when status is changed

MatousJobanek:
* reworked hint to a more elaborated status message 
* status message can be updated 
  * this is used in `test-keeper` where different messages are set for the cases when either the PR is updated by commit containing a test or is updated so it contains only skipped files - for more information see this comment https://github.com/arquillian/ike-prow-plugins/issues/145#issuecomment-397629389
  * for the happy (success) comments, there is used a happy Ike's image 
* the size of Ike's image is smaller
* there is only one possibility of customizing the messages - using `.md` files in `.ike-prow/` directory
  * the files that are checked are:
    * `test-keeper_without_tests_message.md` for the case when no tests is added
    * `test-keeper_with_tests_message.md` for the case when PR is updated by a commit containing a test
    * `test-keeper_only_skipped_message.md` for the case when PR is updated so it contains only those files that are skipped 

Fixes: #148 

TODO:  
- [ ] update documentation


docs/fragments/pr-sanitizer.adoc
```diff
@@ -35,5 +35,7 @@ include::../../pkg/plugin/pr-sanitizer/test_fixtures/github_calls/pr-sanitizer.y
 
 === Status details
 
+Here, you can find status details description applicable for each state of the `pr-sanitizer` plugin.
```
bartoszmajsak: I would replace `Here,` with `In this section`

docs/fragments/work-in-progress.adoc
```diff
@@ -40,5 +40,7 @@ include::../../pkg/plugin/work-in-progress/test_fixtures/github_calls/work-in-pr
 
 === Status details
 
+Here, you can find status details description applicable for each state of the `work-in-progress` plugin.
```
bartoszmajsak: `I would replace `Here,` with `In this section`

pkg/status/message/message_service.go
```diff
@@ -0,0 +1,142 @@
...
+	statusMsg := utils.String("")
+
+	for _, com := range comments {
+		content := *com.Body
+		if strings.HasPrefix(content, s.getPluginTitle()) {
+			s.loadStatusMessage(commentMsgCreator, statusMsg)
+			if strings.TrimSpace(content) == strings.TrimSpace(*statusMsg) {
+				return nil
+			}
+			return s.commentService.EditComment(*com.ID, statusMsg)
```
bartoszmajsak: I'm not sure we should edit comments. I look at PR comments (and events) as a timeline so I would rather append new comments with the relevant message. Maybe we can extend the message to point to commit? "It seems that prior to commit 6fas341213ad3a there are no tests in this PR" or sth like - to get the idea, as the wording is not ideal. WDYT?

pkg/status/message/message_service_test.go
```diff
@@ -113,13 +109,53 @@ var _ = Describe("Config loader features", func() {
...
+			msgService := message.NewStatusMessageService(client, log.NewTestLogger(), commentsLoader, messageContext)
+
+			// when
+			err := msgService.StatusMessage(newBasicMsgCreator("New comment"), true)
+
+			// then
+			Ω(err).ShouldNot(HaveOccurred())
+		})
+
+		It("should modify existing comment when a new message is provided", func() {
```
bartoszmajsak: I am not sure if it is not confusing. We rarely modify our own comments to keep the context of the discussion right, so why would bot do that?

Issue #138, opened 2018-04-27
Lock one version for libs installed by `make tools`


MatousJobanek:
The target `make tools`installs several libraries into `$GOPATH`. The versions of the installed libraries reflect the latest state of the projects. This means that the versions of the libraries can be different in:
* local PR
* travis
* cico jobs that uses docker image

The versions should be locked to one specific commit/tag to keep the behavior consistent

PR #210, opened 2018-06-22
chore(#138): locks version for libraries installed by "make tools".

hemanik:
Locks version for the libraries installed by target `make tools` to provide consistent behavior across local PRs, Travis and cico jobs that use docker images by adding the timestamp to the tools to enable glide to vendor project dependencies.

Fixes: #138


MatousJobanek:
[make tools](https://github.com/arquillian/ike-prow-plugins/blob/master/.make/Makefile.build#L7-L12) is using `go get` command for fetching the tools/libraries, not glide dependency management. The change in `glide.yaml` file has no effect for the tools versions that are being fetched by `go get`

.make/Makefile.build
```diff
@@ -4,13 +4,19 @@ clean: ## Removes binary, cache folder and docker images
...
-tools: ## Installs required go tools
+tools: tools.timestamp
+
+tools.timestamp: ## Installs required go tools
 	@go get -u github.com/alecthomas/gometalinter && gometalinter --install
 	@go get -u github.com/onsi/ginkgo/ginkgo
 	@go get -u github.com/onsi/gomega
 	@go get -u golang.org/x/tools/cmd/goimports
 	@go get -u github.com/jteeuwen/go-bindata/...
+	@touch tools.timestamp
```
bartoszmajsak: What's the purpose for that file?

glide.yaml
```diff
@@ -25,10 +25,15 @@ import:
...
-
 - package: github.com/prometheus/client_golang
   version: faf4ec335fe01ae5a6a0eaa34a5a9333bfbd1a30
-
+- package: github.com/alecthomas/gometalinter
+  version: 46cc1ea3778b247666c2949669a3333c532fa9c6
+- package: github.com/jteeuwen/go-bindata
+  version: 6025e8de665b31fa74ab1a66f2cddd8c0abf887e 
+- package: golang.org/x/tools/cmd/goimports
+  version: 77106db15f689a60e7d4e085d967ac557b918fb2
```
bartoszmajsak: This is not used by our code, so shouldn't be here. 

As `go get` doesn't support versioning, we have two options for `tools` target

- Use http://labix.org/gopkg.in instead of github links
- Use `dep` instead

.make/Makefile.build
hemanik: I intended to store the timestamp and then run `dep` on these as is done for the other `fabric8` repos  - https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/master/Makefile#L66

bartoszmajsak: Can you explain how this file is used? I might be missing something.

hemanik: Sorry, my bad. The purpose of the file is to prevent people from installing the same tools again and again and waste bandwidth by preventing this command if it was run before and we don't need to run it again.

bartoszmajsak: I think we can vote here if we really need this file, so @dipak-pawar @hemanik @lordofthejars @MatousJobanek please :+1: or :-1: this comment.

If we :+1:, then we have to add following:

- add this file to `.gitignore` so we won't accidentally commit it
- probably have a cleanup target

glide.yaml
hemanik: > Use http://labix.org/gopkg.in instead of github links

This fails to install packages. Eg: https://gopkg.in/onsi/gomega.v1

> Use dep instead

Since these packages are not imported directly in the source code. I am unable to specify `-update/-add`  for these specific packages as per the rules described https://golang.github.io/dep/docs/ensure-mechanics.html

I have added these packages in the `Gopkg.toml` with the `[[override]]` and revision. But then this defeats the purpose for having a separate target for the tools. 

@bartoszmajsak @MatousJobanek need your thoughts and suggestions.

hemanik: Or do we write some script that does this-https://stackoverflow.com/posts/32630589/revisions?

bartoszmajsak: This sounds like a bit of overdoing

MatousJobanek: Yeah, I agree with Bartosz. In addition, the script is not complete - you would need to compile the binary and put it to the `/bin` directory.
I was hoping that the `gopkg.in` could work, but it seems that it is not applicable for all cases... But, do we really need to lock the versions for all tools? Looking at it, the most problematic one seems to be the `gometalinter`, so if we are able to lock the version at least for this tool, then I'm fine with it.

hemanik: > But, do we really need to lock the versions for all tools?

Yes, I am not sure how much of an impact an update in tools could have on the code, but I don't think it is very significant unless there is a major change in them. 

If we decide to just lock the version for the `gometalinter`, that is possible using [`gopkg.in`](https://gopkg.in/alecthomas/gometalinter.v2). 
 
@MatousJobanek @bartoszmajsak so what do you propose?

MatousJobanek: I vote for locking version only for `gometalinter` and for keeping the rest of the tools as it is.

hemanik: Updated `gometalinter` and undone all other changes.

MatousJobanek:
[test]

bartoszmajsak:
I think we can leave the unversioned tools, as we can't find a way to lock versions for all the tools.

Issue #127, opened 2018-04-20
Migrate from glide to golang dep


dipak-pawar:
I don't the reason behind why we decided to use `glide` for dependency management, But as the Go community now has the `dep` project to manage dependencies. It would be good to use this(dep) as in near future `glide` will be in maintenance mode instead of active development.

Ref - https://github.com/Masterminds/glide#golang-dep

bartoszmajsak:
> I don't the reason behind why we decided to use glide for dependency management,

The reason is/was that at the moment of creating the project godep was still not the "one and only tool" for dependency management in Go. Also it was familiar for me to get this project up and running very fast as I used it in the past in other go code I wrote. And apart from that - this statement could be a bit more constructive.

Good find with the maintenance mode though - we should migrate at some point, but I don't see it as a high priority, to be honest. OTOH, it's less than 1d of work so we can sneak it in when we have some capacity.

PR #211, opened 2018-06-22
chore(#127): migrate from glide to dep

MatousJobanek:
* replaced glide by dep
* target for updating dependencies is `make ensure`

Fixes: #127

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ before_install:
     fi'
 
 install:
-  - curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+  - curl https://raw.githubusercontent.com/golang/dep/1550da37d8fab9ed2dbc4bd04290e6c8dd3ff04a/install.sh | sh
   - make tools
   - make install
 


.make/Makefile.build
```diff
@@ -12,19 +12,15 @@ tools: ## Installs required go tools
...
-.PHONY: up
-up: generate ## Updates all dependencies defined for glide
-	glide up -v
-
-.PHONY: up-skip-test-quick
-up-skip-test-quick:
-	glide update --skip-test --quick ## Updates only things in glide.yaml and skips tests
+.PHONY: ensure
+ensure: generate ## Updates all dependencies defined for dep
+	dep ensure -update -v
```
bartoszmajsak: Maybe we call it `update` instead? both `dep` targets are using `ensure`, so I was thinking maybe it would be better to name the targets closer to what they do?

setup.sh
```diff
@@ -48,9 +48,9 @@ function install_go_bins() {
       "${LIGHT_GREEN}export GOPATH=\$GOPATH:~/code/golang${CLEAR}"
   fi
 
-  if [ -z $(which glide 2>/dev/null) ]; then
-    echo -e "${CLEAR}${LIGHT_GREEN}Installing glide${CLEAR}"
-    curl https://glide.sh/get | sh
+  if [ -z $(which dep 2>/dev/null) ]; then
+    echo -e "${CLEAR}${LIGHT_GREEN}Installing dep${CLEAR}"
+    curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
```
bartoszmajsak: Are we using `master` or do we want to have it locked to last known working version?

.make/Makefile.build
MatousJobanek: Good point - I'll rename it

setup.sh
MatousJobanek: So, I cannot use the tag because the script wasn't there yet... I'll use the latest commit then...

Issue #201, opened 2018-06-15
Add message hint to PR sanitizer explaining what is missing/wrong


MatousJobanek:
If the PR title/description doesn't comply with the pre-defined structure, then the plugin should add a message hint explaining what is missing/wrong.

MatousJobanek:
Reopening this issue as the PR #213 solves only half of it.

Issue #5, opened 2018-02-26
PR description verification


bartoszmajsak:
- [ ] Simple check if PR content is not too short (let's consider anything shorter than 32 (?) characters not sufficient). If that's the case comment on the PR asking for a more elaborated description.
- [ ] Checks if PR has link to issue so it can  be closed when PR is merged (https://github.com/blog/1506-closing-issues-via-pull-requests)

PR #213, opened 2018-06-26
 feat(#5): verify pr description in pr-sanitizer

dipak-pawar:
*Changes Proposed in this PR*
* Description Length verification - expecting description length should be more than 50 characters excluding issue link and keywords. If a length is too short, commenting on PR for more elaborated description. 
* Issue Link Verification - Expecting at least one issue link as a part of PR description

Fixes: #5, Fixes: #201


pkg/plugin/pr-sanitizer/event_handler.go
```diff
@@ -43,13 +30,51 @@ var (
 	defaultTypes          = []string{"chore", "docs", "feat", "fix", "refactor", "style", "test"}
 )
 
+// DescriptionShortMessage is message notification for contributor about PR short description content.
+const DescriptionShortMessage = "Hey @%s! It seems that PR description is too short. More elaborated description will be helpful to " +
+	"understand changes proposed in this PR."
+
+// HandleEvent is an entry point for the plugin logic. This method is invoked by the Server when
+// events are dispatched from the /hook service
+func (gh *GitHubLabelsEventsHandler) HandleEvent(log log.Logger, eventType github.EventType, payload []byte) error {
```
hemanik: The `HandleEvent` is changed via https://github.com/arquillian/ike-prow-plugins/pull/208. Please update this accordingly.

hemanik: How about renaming `GitHubLabelsEventsHandler` to `GitHubPRSanitizerEventsHandler` or something similar since we do not do anything with the `Labels` here.

pkg/plugin/pr-sanitizer/event_handler.go
```diff
@@ -82,25 +71,53 @@ func (gh *GitHubLabelsEventsHandler) HandleIssueCommentEvent(log log.Logger, com
...
 	config := LoadConfiguration(log, change)
-	if gh.HasTitleWithValidType(config, *pullRequest.Title) {
-		return statusService.Success(SuccessMessage, SuccessDetailsPageName)
-	} else if prefix, ok := wip.GetWorkInProgressPrefix(*pullRequest.Title, wip.LoadConfiguration(log, change)); ok {
-		trimmedTitle := strings.TrimPrefix(*pullRequest.Title, prefix)
-		if gh.HasTitleWithValidType(config, trimmedTitle) {
-			return statusService.Success(SuccessMessage, SuccessDetailsPageName)
+	isTitleWithValidType := gh.HasTitleWithValidType(config, *pr.Title)
+	if !isTitleWithValidType {
+		if prefix, ok := wip.GetWorkInProgressPrefix(*pr.Title, wip.LoadConfiguration(log, change)); ok {
```
bartoszmajsak: What if the particular repository doesn't use WIP plugin and thus considers such prefix as invalid overall? Just a general observation to think through as with this we are coupling both plugins. And I don't really know yet how to address this use-case, so let's think it through.

/cc @MatousJobanek

dipak-pawar: @bartoszmajsak, We are checking first if PR title has required valid type, if not then we are looking for `WIP` in the title, if yes then remove `WIP` prefix and again check if it has required valid type in PR title

docs/fragments/pr-sanitizer.adoc
```diff
@@ -33,9 +33,20 @@ include::../../pkg/plugin/pr-sanitizer/test_fixtures/github_calls/pr-sanitizer.y
...
+=== Description - Content Length Verification [[description-content-length-verification]]
+
+This ensures that the PR description has more than 50 characters to have more elaborated description from which anyone can understand all changes proposed in this PR.
+
+All issue link keywords(e.g. `close #1`) added to close an issue are not taken into consideration while measuring description length.
+
+=== Description - Issue Link Verification [[description-issue-link-verification]]
+
+This ensures that the PR description has issue link using keywords to close issue after merging pull request. Supported keywords for issue links are `closes`, `closed`, `fix`, `fixes`, `fixed`, `resolve`, `resolves`, `resolved`.
+
```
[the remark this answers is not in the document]
dipak-pawar: we are already covering all the cases. I will update plugin behavior if couldn't find issue link

pkg/plugin/pr-sanitizer/event_handler.go
bartoszmajsak: This is exactly what the code does, but that was not what I was wondering about. My question is what if we don't use WIP plugin on a given repo, so if you prefix it with "WIP" it should be considered not correct by this plugin then, right?

In addition it couples both plugins together.

pkg/plugin/pr-sanitizer/pr_sanitizer_message_builder.go
```diff
@@ -0,0 +1,59 @@
...
+
+const (
+	// TitleFailure is a message used in GH Status as description when the PR title does not follow semantic message style
+	TitleFailure = "PR title does not conform with semantic commit message style."
+
+	// IssueLinkMissing is a message used in GH Status as description when no tests shipped with the PR
+	IssueLinkMissing = "Issue link is missing in this PR description."
+
+	// DescriptionLengthShort is a message used in GH Status as description when the PR description does not have minimum required characters.
+	DescriptionLengthShort = "PR description is too short, expecting more than 50 characters."
```
MatousJobanek: These descriptions are great, I also like the API, however, the longer content should be part of the status message, not the status of the builds - the message can be too long and thus hard to read.

pkg/plugin/pr-sanitizer/pr_title_description_status_service.go
```diff
@@ -0,0 +1,40 @@
...
+
+const (
+	// ProwPluginName is an external prow plugin name used to register this service
+	ProwPluginName = "pr-sanitizer"
+
+	// FailureDetailsPageName is a name of a documentation page that contains additional status details for title verification failure.
+	FailureDetailsPageName = "pr-sanitizer-failed"
+
+	// SuccessMessage is a message used in GH Status as description when the PR title conforms to the semantic commit message style
+	SuccessMessage = "PR title conforms with semantic commit message style and description has enough characters with issue link."
```
MatousJobanek: Please, try to make it shorter - the space for the message is not so big

pkg/plugin/pr-sanitizer/pr_title_description_status_service.go
```diff
@@ -0,0 +1,40 @@
+package prsanitizer
+
+import (
+	"github.com/arquillian/ike-prow-plugins/pkg/github"
+	"github.com/arquillian/ike-prow-plugins/pkg/log"
+	"github.com/arquillian/ike-prow-plugins/pkg/scm"
+	"github.com/arquillian/ike-prow-plugins/pkg/status"
+)
+
+type prTitleDescriptionStatusService struct {
```
MatousJobanek: The name of the struct could be shorter - eg: `prSanitizerStatusService`
same for the file: `pr_sanitizer_status_service.go`

pkg/plugin/pr-sanitizer/event_handler.go
```diff
@@ -82,25 +71,53 @@ func (gh *GitHubLabelsEventsHandler) HandleIssueCommentEvent(log log.Logger, com
...
+	if len(failureMessage) > 0 {
+		return statusService.fail(failureMessage)
+	}
+
+	return statusService.titleAndDescriptionOk()
+}
+
+// GetDescriptionWithIssueLinkExcluded return description with excluding issue link keyword.
+func (gh *GitHubPRSanitizerEventsHandler) GetDescriptionWithIssueLinkExcluded(d string) (string, bool) {
+	desc := strings.ToLower(d)
```
MatousJobanek: please, use eg. `description` instead of only `d`

docs/fragments/pr-sanitizer.adoc
```diff
@@ -33,9 +33,25 @@ include::../../pkg/plugin/pr-sanitizer/test_fixtures/github_calls/pr-sanitizer.y
...
+This ensures that the PR description has more than 50 characters to have more elaborated description from which anyone can understand all changes proposed in this PR.
+
+All issue link keywords(e.g. `close #1`) added to close an issue are not taken into consideration while measuring description length.
+
+If it won't find Pull Request description more than 50 characters, it adds a comment with a hint message and marks its check as **Failure**, or as **Success** otherwise.
+
+=== Description - Issue Link Verification [[description-issue-link-verification]]
+
+This ensures that the PR description has issue link using keywords to close issue after merging pull request. Supported keywords for issue links are `closes`, `closed`, `fix`, `fixes`, `fixed`, `resolve`, `resolves`, `resolved`.
+For More info about how can you close issues via PR can be found link:https://blog.github.com/2013-05-14-closing-issues-via-pull-requests[here].
```
MatousJobanek: I would rather point to the description in documentation https://help.github.com/articles/closing-issues-using-keywords/

pkg/plugin/pr-sanitizer/event_handler.go
MatousJobanek: > My question is what if we don't use WIP plugin on a given repo, so if you prefix it with "WIP" it should be considered not correct by this plugin then, right?

Personally, if the title starts with WIP but the rest is a valid semantic title, then I'd like to have the status check green. However, it would be good to have this note in the documentation.

> In addition it couples both plugins together.

You don't have to see this as a coupling of two plugins but as an additional functionality that is brought to this pluing from another source :-)

If you don't like this state, then we can read the `plugins.yaml` file, check if the repository is using work-in-progress plugin and then activate/deactivate this functionality accordingly. That shouldn't be so hard, I just don't see it necessary - but that is my point of view, I don't insist on it ;-)

pkg/plugin/pr-sanitizer/event_handler.go
```diff
@@ -82,25 +71,53 @@ func (gh *GitHubLabelsEventsHandler) HandleIssueCommentEvent(log log.Logger, com
...
+			isTitleWithValidType = gh.HasTitleWithValidType(config, trimmedTitle)
 		}
 	}
-	return statusService.Failure(TitleVerificationFailureMessage, TitleVerificationFailureDetailsPageName)
+
+	description, isIssueLinked := gh.GetDescriptionWithIssueLinkExcluded(pr.GetBody())
+
+	if len(description) < 50 {
+		message := fmt.Sprintf(DescriptionShortMessage, *pr.User.Login)
+		err := gh.Client.CreateIssueComment(scm.RepositoryIssue{
```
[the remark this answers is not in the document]
dipak-pawar: We are just adding a comment here. I believe `StatusMessageService` is required more things than simply using `err := gh.Client.CreateIssueComment(*issue, &message)`

pkg/plugin/pr-sanitizer/event_handler.go
```diff
@@ -82,25 +71,53 @@ func (gh *GitHubLabelsEventsHandler) HandleIssueCommentEvent(log log.Logger, com
...
+		if prefix, ok := wip.GetWorkInProgressPrefix(*pr.Title, wip.LoadConfiguration(log, change)); ok {
+			trimmedTitle := strings.TrimPrefix(*pr.Title, prefix)
+			isTitleWithValidType = gh.HasTitleWithValidType(config, trimmedTitle)
 		}
 	}
-	return statusService.Failure(TitleVerificationFailureMessage, TitleVerificationFailureDetailsPageName)
+
+	description, isIssueLinked := gh.GetDescriptionWithIssueLinkExcluded(pr.GetBody())
+
+	if len(description) < 50 {
```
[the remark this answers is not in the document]
dipak-pawar: we can make it configurable

docs/fragments/pr-sanitizer.adoc
```diff
@@ -32,10 +32,29 @@ include::../../pkg/plugin/pr-sanitizer/test_fixtures/github_calls/pr-sanitizer.y
 
 <1> Defines set of custom types which will be used to match titles prefixes in changeset and determine if Pull Requests Title conforms with the semantic message style. (Eg. above code snippet configures link:https://medium.com/walmartlabs/semantic-commit-messages-with-emojis-dba2541cea9a[emoji style] semantic commit message.)
 <2> Allows you to decide if you want to combine your patterns with the list (`feat`, `fix`, `refactor`, `docs`, `test`, `chore`, `style`) of predefined default types (`true` by default).
+<3> Set Number of characters to be verified in PR description content.
+
+
+=== Description - Content Length Verification [[description-content-length-verification]]
+
+This ensures that the PR description has more than 50 characters to have more elaborated description from which anyone can understand all changes proposed in this PR.
```
hemanik: How about mentioning say 50 characters by default or as configured otherwise?

pkg/plugin/pr-sanitizer/event_handler.go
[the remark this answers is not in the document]
MatousJobanek: `StatusMessageService` is adding a comment as well. It requires more things because it allows us modifying of the comment content as well as avoiding comment duplication. Anyway, that should be part of the #201

Issue #199, opened 2018-06-14
Migrate to Quay


MatousJobanek:
Quay is a new registry OSIO is moving to - specifically https://quay.io/organization/openshiftio
For more information go to the devguide

PR #218, opened 2018-07-02
chore(#199): migrate from devshift registry to quay.io

hemanik:
The plugins and hook services are migrated to the new Quay registry - specifically https://quay.io/organization/openshiftio.

The new repositories for the same are as follows:

CentOS containers:

https://quay.io/openshiftio/ike-prow-plugins-test-keeper
https://quay.io/openshiftio/ike-prow-plugins-pr-sanitizer
https://quay.io/openshiftio/ike-prow-plugins-work-in-progress

RHEL containers:

https://quay.io/openshiftio/rhel-ike-prow-plugins-test-keeper
https://quay.io/openshiftio/rhel-ike-prow-plugins-pr-sanitizer
https://quay.io/openshiftio/rhel-ike-prow-plugins-work-in-progress

Fixes: #199

### DIFF
--- a/.make/Makefile.deploy.prow
+++ b/.make/Makefile.deploy.prow
@@ -21,16 +21,16 @@ build-hook: update
 
 .PHONY: build-hook-image
 build-hook-image:
-	$(DOCKER) build --build-arg BINARY=hook -t $(REGISTRY)/$(DOCKER_REPO)/hook -f $(DEPLOY_DOCKERFILE) .
-	$(DOCKER) tag $(REGISTRY)/$(DOCKER_REPO)/hook $(REGISTRY)/$(DOCKER_REPO)/hook:$(TAG)
+	$(DOCKER) build --build-arg BINARY=hook -t $(REGISTRY)/$(DOCKER_REPO)-hook -f $(DEPLOY_DOCKERFILE) .
+	$(DOCKER) tag $(REGISTRY)/$(DOCKER_REPO)-hook $(REGISTRY)/$(DOCKER_REPO)-hook:$(TAG)
 
 .PHONY: push-hook-image
 push-hook-image: build-hook-image
-	$(DOCKER) push $(REGISTRY)/$(DOCKER_REPO)/hook
+	$(DOCKER) push $(REGISTRY)/$(DOCKER_REPO)-hook
 
 .PHONY: clean-hook-image
 clean-hook-image:
-	$(DOCKER) rmi -f $(REGISTRY)/$(DOCKER_REPO)/hook
+	$(DOCKER) rmi -f $(REGISTRY)/$(DOCKER_REPO)-hook
 
 .PHONY: oc-deploy-plugins ## Builds plugin images, updates configuration and deploys new version of ike-plugins
 oc-deploy-plugins: oc-init-project compile deploy-plugins oc-generate-deployments
@@ -54,16 +54,16 @@ deploy-plugins: build-plugin-images push-plugin-images
 .PHONY: build-plugin-images $(PLUGINS)
 build-plugin-images: $(BUILD_IMAGES)
 $(BUILD_IMAGES): build-%: %
-	$(DOCKER) build --build-arg BINARY=$< -t $(REGISTRY)/$(DOCKER_REPO)/$< -f $(DEPLOY_DOCKERFILE) .
-	$(DOCKER) tag $(REGISTRY)/$(DOCKER_REPO)/$< $(REGISTRY)/$(DOCKER_REPO)/$<:$(TAG)
+	$(DOCKER) build --build-arg BINARY=$< -t $(REGISTRY)/$(DOCKER_REPO)-$< -f $(DEPLOY_DOCKERFILE) .
+	$(DOCKER) tag $(REGISTRY)/$(DOCKER_REPO)/$< $(REGISTRY)/$(DOCKER_REPO)-$<:$(TAG)
 
 
 .PHONY: clean-plugin-images
 clean-plugin-images: $(CLEAN_IMAGES)
 $(CLEAN_IMAGES): clean-%: %
-	$(DOCKER) rmi -f $(REGISTRY)/$(DOCKER_REPO)/$<
+	$(DOCKER) rmi -f $(REGISTRY)/$(DOCKER_REPO)-$<
 
 .PHONY: push-plugin-images
 push-plugin-images: build-plugin-images $(PUSH_IMAGES)
 $(PUSH_IMAGES): push-%: %
-	$(DOCKER) push $(REGISTRY)/$(DOCKER_REPO)/$<
+	$(DOCKER) push $(REGISTRY)/$(DOCKER_REPO)-$<
--- a/cico_setup.sh
+++ b/cico_setup.sh
@@ -11,7 +11,7 @@ set -e
 function load_jenkins_vars() {
   if [ -e "jenkins-env" ]; then
     cat jenkins-env \
-      | grep -E "(DEVSHIFT_TAG_LEN|DEVSHIFT_USERNAME|DEVSHIFT_PASSWORD|JENKINS_URL|GIT_BRANCH|GIT_COMMIT|BUILD_NUMBER|ghprbSourceBranch|ghprbActualCommit|BUILD_URL|ghprbPullId)=" \
+      | grep -E "(DEVSHIFT_TAG_LEN|QUAY_USERNAME|QUAY_PASSWORD|JENKINS_URL|GIT_BRANCH|GIT_COMMIT|BUILD_NUMBER|ghprbSourceBranch|ghprbActualCommit|BUILD_URL|ghprbPullId)=" \
       | sed 's/^/export /g' \
       > ~/.jenkins-env
     source ~/.jenkins-env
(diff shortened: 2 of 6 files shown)


bartoszmajsak:
[test]

.make/Makefile.deploy.prow
```diff
@@ -21,16 +21,16 @@ build-hook: update
 
 .PHONY: build-hook-image
 build-hook-image:
-	$(DOCKER) build --build-arg BINARY=hook -t $(REGISTRY)/$(DOCKER_REPO)/hook -f $(DEPLOY_DOCKERFILE) .
-	$(DOCKER) tag $(REGISTRY)/$(DOCKER_REPO)/hook $(REGISTRY)/$(DOCKER_REPO)/hook:$(TAG)
+	$(DOCKER) build --build-arg BINARY=hook -t $(REGISTRY)/$(DOCKER_REPO)-hook -f $(DEPLOY_DOCKERFILE) .
+	$(DOCKER) tag $(REGISTRY)/$(DOCKER_REPO)-hook $(REGISTRY)/$(DOCKER_REPO)-hook:$(TAG)
```
bartoszmajsak: Why is it `-` instead of `/` now?

cico_setup.sh
```diff
@@ -49,17 +49,17 @@ function cleanup_env {
...
     export DEPLOY_DOCKERFILE='Dockerfile.deploy.rhel'
-    export DOCKER_REPO="osio-prod/ike-prow-plugins"
+    export DOCKER_REPO="openshiftio/rhel-ike-prow-plugins"
   fi
 
   # Login first
-  if [ -n "${DEVSHIFT_USERNAME}" -a -n "${DEVSHIFT_PASSWORD}" ]; then
-    docker login -u ${DEVSHIFT_USERNAME} -p ${DEVSHIFT_PASSWORD} ${REGISTRY}
+  if [ -n "${QUAY_USERNAME}" -a -n "${QUAY_PASSWORD}" ]; then
+    docker login -u ${QUAY_USERNAME} -p ${QUAY_PASSWORD} ${REGISTRY}
```
bartoszmajsak: I would suggest using `IMG_REPO` prefix instead of `QUAY`. This will then have a proper meaning for any repo we use.

.make/Makefile.deploy.prow
bartoszmajsak: Maybe we can also change `DOCKER_REPO` -> `IMG_REPO`?

hemanik: Because it changed from `prod.registry.devshift.net/osio-prod/ike-prow-plugins/test-keeper` to `quay.io/openshiftio/ike-prow-plugins-test-keeper` as per the new repositories. 

Here's the discussion -  https://chat.openshift.io/developers/pl/tqzbcjx5f3nmdg7co9fn5kay5a

cluster/hook-template.yaml
```diff
@@ -42,7 +42,7 @@ objects:
           terminationGracePeriodSeconds: 180
           containers:
           - name: hook
-            image: ${REGISTRY}/${DOCKER_REPO}/hook:${IMAGE_TAG}
+            image: ${REGISTRY}/${DOCKER_REPO}-hook:${IMAGE_TAG}
```
bartoszmajsak: I don't think this is correct now. What is the value here for the generated `yaml` file?

.make/Makefile.deploy.prow
bartoszmajsak: Thanks for the link. In such a case I think we should improve the naming here, as ` $(REGISTRY)/$(DOCKER_REPO)-hook` is not really correct. 

In our case this should be:

`quay.io/openshiftio/rhel-ike-prow-plugins-work-in-progress`

so 

 * `REGISTRY = quay.io`
 * `IMG_REPO = openshift.io`
*  `ike-prow-plugins-` is a prefix of the plugin image name in such case.

hemanik:
@jmelis can you please verify this PR mainly from the centos-ci publishing point of view?

.make/Makefile.deploy.prow
```diff
@@ -21,16 +22,16 @@ build-hook: update
 
 .PHONY: build-hook-image
 build-hook-image:
-	$(DOCKER) build --build-arg BINARY=hook -t $(REGISTRY)/$(DOCKER_REPO)/hook -f $(DEPLOY_DOCKERFILE) .
-	$(DOCKER) tag $(REGISTRY)/$(DOCKER_REPO)/hook $(REGISTRY)/$(DOCKER_REPO)/hook:$(TAG)
+	$(DOCKER) build --build-arg BINARY=hook -t $(REGISTRY)/$(IMG_REPO)/$(OC_PROJECT_NAME)-hook -f $(DEPLOY_DOCKERFILE) .
+	$(DOCKER) tag $(REGISTRY)/$(IMG_REPO)-hook $(REGISTRY)/$(IMG_REPO)/$(OC_PROJECT_NAME)-hook:$(TAG)
```
bartoszmajsak: Shouldn't `$(REGISTRY)/$(IMG_REPO)-hook` be `$(REGISTRY)/$(IMG_REPO)/$(OC_PROJECT_NAME)-hook` now?

cico_setup.sh
```diff
@@ -11,7 +11,7 @@ set -e
 function load_jenkins_vars() {
   if [ -e "jenkins-env" ]; then
     cat jenkins-env \
-      | grep -E "(DEVSHIFT_TAG_LEN|DEVSHIFT_USERNAME|DEVSHIFT_PASSWORD|JENKINS_URL|GIT_BRANCH|GIT_COMMIT|BUILD_NUMBER|ghprbSourceBranch|ghprbActualCommit|BUILD_URL|ghprbPullId)=" \
+      | grep -E "(IMG_REPO_TAG_LEN|IMG_REPO_USERNAME|IMG_REPO_PASSWORD|JENKINS_URL|GIT_BRANCH|GIT_COMMIT|BUILD_NUMBER|ghprbSourceBranch|ghprbActualCommit|BUILD_URL|ghprbPullId)=" \
```
bartoszmajsak: I don't know from where `DEVSHIFT_*` is coming. Might be env variable of Centos-CI. Can you make sure we are doing the right thing here?

jmelis: That is correct, `DEVSHIFT_USERNAME` and `DEVSHIFT_PASSWORD` are coming from CentOS-CI, however if migrating to quay, we should be using `QUAY_USERNAME` and `QUAY_PASSWORD` instead. Can you gobally replace `DEVSHIFT_USERNAME` and `DEVSHIFT_PASSWORD` with their `quay` counterparts? also, please keep `DEVSHIFT_TAG_LEN` as is, that's  a special variable that contains the expected length of git commit hash and can't be renamed.

Also, I recommend you to no longer use such a mechanismo to load the env, but rather use this:

```bash
function load_jenkins_vars() {
  if [ -e "jenkins-env.json" ]; then
    eval "$(./env-toolkit load -f jenkins-env.json \
        DEVSHIFT_TAG_LEN \
        QUAY_USERNAME \
        QUAY_PASSWORD \
        JENKINS_URL \
        GIT_BRANCH \
        GIT_COMMIT \
        BUILD_NUMBER \
        ghprbSourceBranch \
        ghprbActualCommit \
        BUILD_URL \
        ghprbPullId)"
  fi
}
```

It's a much safer approach as it's resilient to env vars with spaces, newlines and special chars.

cico_setup.sh
```diff
@@ -49,17 +49,17 @@ function cleanup_env {
...
     export DEPLOY_DOCKERFILE='Dockerfile.deploy.rhel'
-    export DOCKER_REPO="osio-prod/ike-prow-plugins"
+    export IMG_REPO="openshiftio/rhel-"
   fi
 
   # Login first
-  if [ -n "${DEVSHIFT_USERNAME}" -a -n "${DEVSHIFT_PASSWORD}" ]; then
-    docker login -u ${DEVSHIFT_USERNAME} -p ${DEVSHIFT_PASSWORD} ${REGISTRY}
+  if [ -n "${IMG_REPO_USERNAME}" -a -n "${IMG_REPO_PASSWORD}" ]; then
+    docker login -u ${IMG_REPO_USERNAME} -p ${IMG_REPO_PASSWORD} ${REGISTRY}
```
jmelis: please replace with `QUAY_...`

.make/Makefile.deploy.prow
jmelis: I think so too.

hemanik:
> You need to check for the TARGET env variable and check if $TARGET = rhel or not.

[Here,](https://github.com/arquillian/ike-prow-plugins/pull/218/files#diff-66536a55e28a60c7e24d2d4a43a8b89cR62) we check if the `$TARGET=rhel` and the value is overriden for RHEL based containers otherwise the value is fetched from the [Makefile](https://github.com/arquillian/ike-prow-plugins/pull/218/files#diff-b67911656ef5d18c4ae36cb6741b7965R12). 

@jmelis @bartoszmajsak, correct me if this is incorrect and we need to explictly check for the `CentOS`  target in the cico_setup.sh script.

hemanik:
Thanks @jmelis @bartoszmajsak for your review. I have incorporated the suggested changes.

MatousJobanek:
[test]

.make/Makefile.deploy.prow
```diff
@@ -54,16 +56,16 @@ deploy-plugins: build-plugin-images push-plugin-images
 .PHONY: build-plugin-images $(PLUGINS)
 build-plugin-images: $(BUILD_IMAGES)
 $(BUILD_IMAGES): build-%: %
-	$(DOCKER) build --build-arg BINARY=$< -t $(REGISTRY)/$(DOCKER_REPO)/$< -f $(DEPLOY_DOCKERFILE) .
-	$(DOCKER) tag $(REGISTRY)/$(DOCKER_REPO)/$< $(REGISTRY)/$(DOCKER_REPO)/$<:$(TAG)
+	$(DOCKER) build --build-arg BINARY=$< -t $(REGISTRY)/$(IMG_REPO)/$(OC_PROJECT_NAME)-$< -f $(DEPLOY_DOCKERFILE) .
+	$(DOCKER) tag $(REGISTRY)/$(IMG_REPO)/$(OC_PROJECT_NAME)-$< $(REGISTRY)/$(IMG_REPO)/$(OC_PROJECT_NAME)-$<:$(TAG)
```
MatousJobanek: Am I missing something or does the combination of:
https://github.com/arquillian/ike-prow-plugins/pull/218/files#diff-66536a55e28a60c7e24d2d4a43a8b89cR64
`export IMG_REPO="openshiftio/rhel-"`
with 
`$(REGISTRY)/$(IMG_REPO)/$(OC_PROJECT_NAME)-$<`
result in
`quay.io/openshiftio/rhel-/ike-prow-plugins-test-keeper`
which is not correct

bartoszmajsak: @hemanik Is there a fix coming for it though? I can't see updates and would love to close this off.

hemanik: @bartoszmajsak, will update it.

PR #219, opened 2018-07-10
fix(#201): add status message to pr-sanitizer

### LINKED ISSUE #201: Add message hint to PR sanitizer explaining what is missing/wrong

MatousJobanek:
some work to the issue #201 has been already done in the PR #213 but the implementation wasn't complete - the missing bits that are provided in this PR are:
* more customized content of the status messages 
    * contains the list of semantic prefixes valid for the particular repo
    * contains the expected and actual length of the description
* when the PR is good (plugin check is green) then the status message is updated
* brings support for customizing the status messages

apart from that this PR also
* slightly rewrites the docs
* introduces abstraction of checks to the `pr-sanitizer` - every check validates one thing (title, desc length, and issue link) and returns a formatted failure message if the check fails
* refactors tests
* fixes check of the semantic title prefix - originally it was possible to have `fixes failing test` as the title and the check was green

the comment (when all checks fail) looks like:
obrazek

when the PR is updated so it is only missing the issue link:
obrazek

and when everything is good:
obrazek

Fixes #201


bartoszmajsak:
`alien-ike` is dead right now, we will have to trigger it manually once the cluster is healthy

docs/fragments/pr-sanitizer.adoc
```diff
@@ -2,9 +2,11 @@
...
 
+If any of the `pr-sanitizer` checks finds out that the pull request doesn't comply with any of the predefined conventions then it marks the plugin's check as **Failure**, or as **success** otherwise.
+
 === Title Verification [[title-verification]]
 
-This ensures that the PR title comforms with the semantic commit message style. Conformance with the semantic commit message style not only makes your changelog and git history clean and easy to navigate but also encourages making atomic pull requests that address only a single concern, better reflect the change within and are more clearly understood.
+This ensures that the PR title conforms with the semantic commit message style. Conformance with the semantic commit message style not only makes your changelog and git history clean and easy to navigate but also encourages making atomic pull requests that address only a single concern, better reflect the change within and are more clearly understood.
```
bartoszmajsak: How about a link to that here too? We can make an alias for `https://seesparkbox.com/foundry/semantic_commit_messages` and link whenever we say "semantic commit message".

pkg/plugin/pr-sanitizer/event_handler_unit_test.go
```diff
@@ -33,45 +30,67 @@ var _ = Describe("PR-Sanitizer Plugin features", func() {
...
-		BeforeEach(func() {
-			handler = &prsanitizer.GitHubPRSanitizerEventsHandler{Client: NewDefaultGitHubClient()}
-		})
+		DescribeTable("should NOT recognize issue link presence",
+			func(desc string) {
+				pr := &github.PullRequest{Body: utils.String(desc)}
+				msg := prsanitizer.CheckIssueLinkPresence(pr, prsanitizer.PluginConfiguration{}, log.NewTestLogger())
+				Expect(msg).NotTo(BeEmpty())
+			},
+			Entry("missing hash sign", "PR\r\n\r\nfixes: 1 issue"),
```
bartoszmajsak: Might be my obsession, but can we make some tests with uppercase keywords too?

MatousJobanek: I like this obsession :-P 6369164

docs/fragments/pr-sanitizer.adoc
MatousJobanek: I added the link for every use case when we mention the semantic commit message:

* general description
* failure status details page
* status message comment

but only once for every page - I tried to use it every time where the  "semantic commit message" phrase is used, but it looks a bit weird - referencing the same source several times in the same paragraph.

7ee236a

